PR #11, opened 2018-04-14
default-theme: use v-html span for feature.details

LewisJEllis:
I wanted to include links in the feature details, but they're rendered using standard mustaches so my links were escaped. Switching to a v-html span lets me include links, and I think for this purpose it's fair to trust the input of feature details not to be escaped.

My current workaround is to have my package.json look like this:
```
  "scripts": {
    "vuepress-feature-html-hack": "sed -i -e 's|{{ feature.details }}|<span v-html=\"feature.details\"></span>|g' node_modules/vuepress/lib/default-theme/Home.vue",
    "build-docs": "npm run vuepress-feature-html-hack && vuepress build docs",
```
Very hacky, but feels better than duplicating the entire default theme except this one part.

### DIFF
--- a/lib/default-theme/Home.vue
+++ b/lib/default-theme/Home.vue
@@ -15,7 +15,7 @@
     <div class="features">
       <div class="feature" v-for="feature in data.features">
         <h2>{{ feature.title }}</h2>
-        <p>{{ feature.details }}</p>
+        <p><span v-html="feature.details"></span></p>
       </div>
     </div>
     <Content custom/>


egoist:
Maybe render `details` as Vue template as well? so that we can use `<router-link>` to do navigations there 😂

yyx990803:
I'd avoid `v-html` as much as we can, I'll think about how to make it more flexible. Maybe we should extract the features content from markdown instead of the frontmatter.

LewisJEllis:
Makes sense - I did intuitively try a markdown link first before trying HTML, and I do think that would make for a better/more consistent experience. Would save from having to include my own `rel="noopener"`, `target="_blank"` etc attributes too.

Let me know what ideas you come up with and if it's not too complicated to implement I can try updating this PR.

yyx990803:
See https://github.com/vuejs/vuepress/issues/101#issuecomment-382109175

Not using `v-html` because it won't be able to include router links or Vue components.

PR #23, opened 2018-04-14
Allow index.md files as index pages

jaylinski:
Until now, only `README.md` files got treated as a directory index.

This commit also allows `index.md` to be used as index files.

Related to #17.

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -189,7 +189,7 @@ async function genComponentRegistrationFile ({ sourceDir }) {
   return `import Vue from 'vue'\n` + components.map(genImport).join('\n')
 }
 
-const indexRE = /\breadme\.md$/i
+const indexRE = /\b(index|readme)\.md$/i
 const extRE = /\.(vue|md)$/
 
 function fileToPath (file) {


filipsobol:
Hmm... I like this, but wouldn't it be nice if index page was configurable. Something like "index" or "entry" key in config.js?

yyx990803:
I think index/readme are good enough. Sometimes fewer options is better...

emrahcetiner:
> I think index/readme are good enough. Sometimes fewer options is better...

index/readme md file rendered on gitlab/github repository home page by default.
i want to use readme.md for project readme file, use home.md for vuepress home/index.

@filipsobol 's suggestion is good! vuepress index page can be configurable in config.js, like index key.

alancwoo:
I'm confused, is there a way to use `index.md` as the default content for the landing page, so that `readme.md` can exist and be used only for the git repo? My site seems to prioritize `readme.md` over `index.md`

veidenberg:
> is there a way to use `index.md` as the default content for the landing page

The default file seems to be chosen here:

https://github.com/vuejs/vuepress/blob/5023d195dcd7ee6d3710a43558940bcdc31d5a2f/lib/prepare.js#L239

For some reason, VuePress uses `index.md` over `readme.md` in dev mode, but the other way around in build mode.

> i want to use readme.md for project readme file, use home.md for vuepress home/index.

Rename the GitHub readme as `README.markdown` and Vuepress file as `index.md` or `readme.md`.

fPkX6F1nGTX:
I am confused reading this thread in 2024. Does Github allow people to rename README.md to index.md and it will still display on the front of the repo to all visitors?

fPkX6F1nGTX:
> I am confused reading this thread in 2024. Does Github allow people to rename README.md to index.md and it will still display on the front of the repo to all visitors?

Also if anyone cares to contribute to the discussion on SO, please do so:
https://stackoverflow.com/a/69190614/12763497

Issue #26, opened 2018-04-14
When port 8080 is in use nothing happened


FadySamirSadek:
Expected behaviour   :The cli should prompt me if certain port is already in use and suggest another port.
If this is a confirmed issue I would love to pick it up 👍

yyx990803:
PR welcome.

PR #30, opened 2018-04-14
fix(dev build): reserving unused port #26

FadySamirSadek:


lib/dev.js
```diff
@@ -71,7 +72,12 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
   }
 
   const compiler = webpack(config)
-  const port = cliOptions.port || options.siteConfig.port || 8080
+  portfinder.basePort = cliOptions.port || options.siteConfig.port || 8080
+  const port = await portfinder.getPortPromise()
+    .then((port) => {
+      return port
```
ulivz: .then here can be removed if you only want to get the available port.

lib/dev.js
```diff
@@ -71,7 +72,9 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
   }
 
   const compiler = webpack(config)
-  const port = cliOptions.port || options.siteConfig.port || 8080
+  portfinder.basePort = cliOptions.port || options.siteConfig.port || 8080
+  const port = await portfinder.getPortPromise()
+    .catch(err => console.log(err))
```
yyx990803: no need to catch here, this should propagate up the call stack and be logged anyway.

PR #38, opened 2018-04-15
Update dev instruction to yarn

ulivz:
Since vuepress uses yarn instead of npm, so it's better to recommend contributors to use yarn instead of generating npm's locked file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ Contributions are welcome!
 ## Development
 
 ``` bash
-npm install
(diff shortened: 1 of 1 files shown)


filipsobol:
I think we should have both, simply because most of the users still use npm.

`
# using NPM
npm install
npm run dev # serves VuePress docs

# using Yarn
yarn
yarn dev # serves VuePress docs
`

PR #43, opened 2018-04-15
Clarify the description in Browser API Access Restrictions.

ulivz:
The `Browser API Access Restrictions` is only existing before generating static builds. make this description clearly.

### DIFF
--- a/docs/guide/using-vue.md
+++ b/docs/guide/using-vue.md
@@ -2,7 +2,7 @@
 
 ## Browser API Access Restrictions
 
-Because VuePress applications are server-rendered in Node.js when generating static builds, any Vue usage must conform to the [universal code requirements](https://ssr.vuejs.org/en/universal.html). In short, make sure to only access Browser / DOM APIs in `beforeMounted` or `mounted` hooks.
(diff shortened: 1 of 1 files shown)


docs/guide/using-vue.md
```diff
@@ -2,7 +2,7 @@
 
 ## Browser API Access Restrictions
 
-Because VuePress applications are server-rendered in Node.js when generating static builds, any Vue usage must conform to the [universal code requirements](https://ssr.vuejs.org/en/universal.html). In short, make sure to only access Browser / DOM APIs in `beforeMounted` or `mounted` hooks.
+Because VuePress applications are server-rendered in Node.js before generating static builds, any Vue usage must conform to the [universal code requirements](https://ssr.vuejs.org/en/universal.html). In short, make sure to only access Browser / DOM APIs in `beforeMounted` or `mounted` hooks.
```
egoist: `when` sounds better to me 😅

ulivz: When I understand `generating static builds` is for `dev` mode, I also think `when` is better. just my first instinct is for `build`. 😂

PR #61, opened 2018-04-15
Allow injecting Vue

fjc0k:
This PR is helpful for component developers to show their components, e.g.

`.vuepress/inject.js`:

```js
import FirUI from '../../dist/fir-ui.min.js'
import '../../dist/fir-ui.min.css'

export default Vue => {
  Vue.use(FirUI)
}
```

`components/button.md`:

```md
<f-button type="primary">Button</f-button>
```

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -24,9 +24,11 @@ module.exports = async function prepare (sourceDir) {
   // 2. generate routes & user components registration code
   const routesCode = await genRoutesFile(options)
   const componentCode = await genComponentRegistrationFile(options)
+  const injectCode = await genInjectFile(options)
 
   await writeTemp('routes.js', [
     componentCode,
+    injectCode,
     routesCode
   ].join('\n\n'))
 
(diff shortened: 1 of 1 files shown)


ulivz:
Once the new version has supported to eject the default theme (89538fa92d6fb2c0b438df9df2554c9938a81d93) which can let you to mutate the Vue constructor, so I'm not sure whether it's necessary. Maybe this PR is good for those users who don't want to eject theme but still want to extend Vue.

fjc0k:
@ulivz The new feat EJECT is good to develop a new theme. But most of the time, we just need to extend Vue. What's more, we want keep it independent on THE THEME.

lib/prepare.js
```diff
@@ -250,6 +252,17 @@ async function genRoutesFile ({ siteData: { pages }, sourceDir, pageFiles }) {
...
+  const injectFile = path.resolve(sourceDir, '.vuepress/inject.js')
+  if (!fs.existsSync(injectFile)) {
+    return
+  }
+  return (
+    `import inject from ${JSON.stringify(injectFile)}\n` +
+    `inject(Vue)`
+  )
+}
+
```
egoist: maybe adding a webpack alias to this file and import it in `lib/app/app.js` would be better

ulivz: IMO, This solution would be better so don't need to open a new API.

egoist:
And there's no need to inject `Vue` as the argument since you can directly import it, something else like `app` instance may be be more useful.

fjc0k:
#80

PR #115, opened 2018-04-18
feat:support yaml config

Tsuki:
Add support yaml config format


Tsuki:
Maybe split the base config and the advanced (Type: Function)
Most of the config can cover by yaml format.

PR #130, opened 2018-04-18
fix: tip after the server has started

wangyi7099:


lib/dev.js
```diff
@@ -82,7 +82,7 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
       isFirst = false
       console.log(
         `\n  VuePress dev server listening at ${
-          chalk.cyan(`http://localhost:${port}${options.publicPath}`)
+          chalk.cyan(`http://${cliOptions.host || options.siteConfig.host || '0.0.0.0'}:${port}${options.publicPath}`)
```
ulivz: Why setting the fallback to `'0.0.0.0'` instead of `'localhost'`?

wangyi7099: @ulivz 

Because I see the following host setting of the context fallback to  `localhost`

PR #136, opened 2018-04-18
Update slugify.js to remove special quotes

ycmjason:
I have added `“”’` to the `rSpecial` regex.

This is to fix that the sidebar wouldn't work for titles with quotes. This is due to markdown-it is compiling quotes to those special characters and the our `slugify` function do not remove them.

From the above photo, my mouse is hovering on the `#` and as you can see it is directing me to `http://localhost:8082/blog/2018-04-18.html#the-%E2%80%9Chello-world%E2%80%9D-website`, yet the sidebar directs me to `http://localhost:8082/blog/2018-04-18.html#the-hello-world-website`.

### DIFF
--- a/lib/markdown/slugify.js
+++ b/lib/markdown/slugify.js
@@ -3,7 +3,7 @@
 const removeDiacritics = require('diacritics').remove
 // eslint-disable-next-line no-control-regex
 const rControl = /[\u0000-\u001f]/g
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks, this has already been fixed by 5f7e199

ycmjason:
@yyx990803  why do we want to remove the other characters? Shouldn't we remove all special chars to make a cleaner url?

PR #138, opened 2018-04-18
feat: support toml config

lbenie:
[#140]


jaylinski:
What's the use-case for this, or: what's so bad about a js config?

Supporting all kinds of formats for a config file sounds like feature creep for me... (same for #115)

yyx990803:
I think it's just users coming from other static site generators prefer some of these more popular formats. I think we'll support js, yaml and toml and that's it.

yyx990803:
@lbenie I can verify the code is working with the preview - ccould you please delete the toml version of the config before I merge?

lbenie:
@yyx990803 sure thing

PR #148, opened 2018-04-19
Added reference to emoji list

ycmjason:


docs/guide/markdown.md
```diff
@@ -79,6 +79,8 @@ meta:
 
 :tada: :100:
 
+[A list of all emojis available can be found here.](https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/full.json)
```
ulivz: Only linked "here" would looks better:

```md
A list of all emojis available can be found [here](https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/full.json).
```

A list of all emojis available can be found [here](https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/full.json).

ycmjason:
@ulivz fixed

Issue #36, opened 2018-04-15
Blog Support roadmap


chuckdries:
Hello,

I'm excited to use vuepress for my personal website but I kinda want blog support before I switch, so I was thinking I'd kickstart that discussion on that now. This issue is to seek input as to what blog support might look like.

I was thinking a minimum viable product might be
 - [ ] some kind of tag/category system
 - [ ] An alternative default theme which feels more like a blog
 - [ ] helper utilities for themes (high order components?) for getting lists of posts around the tag/category system
 - [ ] Pre-defined markdown templates that correspond to theme vue components, like hexo's layouts
 - [ ] A CLI command for bootstrapping posts, `vuepress new [layout] <post title>`

FadySamirSadek:
I think you should add also the functionality to publish only posts with dates that has already came like gatsby and jekyll. If you need help I would absolutely love to work on these tasks with you

gaomd:
Also add the RSS/JSON Feed and Draft.

mdaffin:
The ability to alias posts and create redirects for them to allow movement of old posts while preserving all incoming links.

ChinuxParibus:
Maybe list all post in some sort of filter (customizable, of course), tags or categories and show them in README.md at first.

lekevicius:
I have been reading VuePress code and comparing with the blog build system I have written for Gulp.

List of features I have for [my gulp static blog](https://github.com/lekevicius/lekevicius.com/blob/master/gulpfile.coffee):

- [x] Markdown
- [x] Custom frontmatter
- [x] Per-post CSS and JS
- [ ] Content in siteData to create feed, index page, etc.
- [ ] RSS/JSON feed
- [ ] Tags in frontmatter
- [ ] Date parsing from filename
- [ ] Virtual directories in URL, as described by @gaomd, turning `2018-04-16-post-title.md` to `blog/2018/04/16/post-title/index.html`
- [ ] Custom helper functions, like reading time calculation from markdown
- [ ] Pagination for blog page
- [ ] Paginated monthly archives
- [ ] Paginated tag pages

These are all the features I've built for gulp static site building, and would like VuePress to have.

However, adding all of that to the "core" would not be very great. I think a better, smarter solution would be to think of a way to allow themes to modify/plug into `prepare` and `build` stages, creating virtual pages, feed, etc.

I'm happy to help building and designing Blog theme, but I'd like to work out this build stage plugins before, instead of forking and modifying "core".

[...]

PR #154, opened 2018-04-19
Theme index enhancment support

mdaffin:
Adds support for loading index.js in themes.

This works in the same way that `.vuepress/enhanceApp.js` works but for themes instead, allowing theme designers to extend the vue app in more useful ways.

This should help with creating custom themes for things like blogging sites like requested in #36

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -52,24 +52,21 @@ if (!Object.assign) Object.assign = require('object-assign')`
     await writeTemp(`override.styl`, hasUserOverride ? `@import(${JSON.stringify(overridePath)})` : ``)
   }
 
+  async function writeEnhanceTemp (destName, srcPath, isEnhanceExist) {
+    await writeTemp(
+      destName,
+      isEnhanceExist
+        ? `export { default } from ${JSON.stringify(srcPath)}`
+        : `export default function () {}`
+    )
+  }
+
   // 6. handle enhanceApp.js
-  const enhancePath = path.resolve(sourceDir, '.vuepress/enhanceApp.js')
-  const hasEnhancePath = fs.existsSync(enhancePath)
-  await writeTemp(
-    'enhanceApp.js',
-    hasEnhancePath
-      ? `export { default } from ${JSON.stringify(enhancePath)}`
-      : `export default function () {}`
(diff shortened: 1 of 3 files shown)


lib/prepare.js
```diff
@@ -62,6 +62,15 @@ if (!Object.assign) Object.assign = require('object-assign')`
...
+  // 7. handle the theme index.js
+  const hasThemeIndexPath = fs.existsSync(options.themeApp)
+  await writeTemp(
+    'themeApp.js',
+    hasThemeIndexPath
+      ? `export { default } from ${JSON.stringify(options.themeApp)}`
+      : `export default function () {}`
+  )
+
   return options
```
ulivz: `themeApp ` and `enhanceApp` are the same except  `hasEnhancePath` and `hasThemeIndexPath`, which can be extracted.

mdaffin: Not exactly the same, they are two proxy files to two different js files that may or may not exist. They will default to an empty function if they don't exist but are there to load in two different files `<theme>/index.js` and `.vuepress/enchanceApp.js` respectively into the main vue app.

ulivz: hmmm... you misunderstand my comment, I know WHY and HOW you do that...
I just want you to refactor your code like this:
```js
  async function writeEnhanceTemp (destName, srcPath, isEnhanceExist) {
    await writeTemp(
      destName,
      isEnhanceExist
        ? `export { default } from ${JSON.stringify(srcPath)}`
        : `export default function () {}`
    )
  }

  // 6. handle enhanceApp.js
  const enhanceAppPath = path.resolve(sourceDir, '.vuepress/enhanceApp.js')
  writeEnhanceTemp('enhanceApp.js', enhanceAppPath, fs.existsSync(enhanceAppPath))

  // 7. handle the theme index.js
  writeEnhanceTemp('themeApp.js', options.themeApp, fs.existsSync(options.themeApp))
```

mdaffin: Ah, ok, done

mdaffin:
Before releasing this we should consider if we want to stick with `index.js` or use something else as we might want to use `index.js` for theme plugins as in #196

PR #158, opened 2018-04-19
SiteMap option

dgpgdev:
adding sitemap file generator on build .

to enabled just add a new config property in config file 

```javascript
  dest: 'vuepress',
  sitemap: {
    filename: 'sitemap.txt',
    domain: 'http://test.com'
  },
  base:'/base-doc/',
```
domain are exact domain name to host doc and filename. When you build your files (build, not dev)

Now you can generate txt or xml sitemap. Check extension file from filename property in filename.
By default if filename is empty a sitemap.txt will be created.


dgpgdev:
adding a new property in objectfile (prepare.js)

lastModified  = date iso form 

and added lastmod tag in xml sitemap

dgpgdev:
lastModified can be use to see page updated recently : 

capture du 2018-04-20 13-13-26

Issue #141, opened 2018-04-18
feat: support toml front matter


slathrop:
## Similar to other SSGs ([Hugo](https://gohugo.io/content-management/front-matter/), etc.), add support in VuePress for TOML in addition to YAML.

- [ ] Select or write package equivalent to yaml-front-matter
- [ ] Add reference to /lib/prepare.js
- [ ] Complete implementation
- [ ] Complete Testing

yyx990803:
Would be open to a PR that switch to `gray-matter`.

colinbate:
The current integration of `gray-matter` is a good start, but it doesn't support the more common TOML frontmatter delimiters of `+++`. Unfortunately, `gray-matter` doesn't implement this auto-detection, but [`preliminaries`](https://github.com/josephearl/preliminaries) does. I'm not suggesting another switch, as it should be something that VuePress can implement itself. At least until it gets added upstream.

If I get a chance in the next couple of days, I should be able to make a PR.

PR #164, opened 2018-04-20
feat: support for TOML front matter (#141)

ryanriv:
I took a stab at swapping out yaml-front-matter for gray-matter to enable TOML front matter. In addition, JSON front matter also works (but I believe it did with yaml-front-matter as well).

Here's an example of how to use TOML front-matter:

```
---toml
title = 'This is a title'
[[meta]]
  name = 'description'
  content = 'hello'
---
```

I also refactored the front matter parsing code a bit, making use of the existing `parseFrontmatter` util.

P.S. This is my first pull request. Please let me know if I am missing anything 😄

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -2,10 +2,9 @@ const path = require('path')
 const fs = require('fs-extra')
 const globby = require('globby')
 const yamlParser = require('js-yaml')
-const yaml = require('yaml-front-matter')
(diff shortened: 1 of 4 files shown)


yyx990803:
Great job for a 1st PR!

ulivz:
@robmuh Welcome to open a PR for that. Features added in the early days may have forgotten to add doc, but should be rare.

Issue #167, opened 2018-04-20
Repo link is hardcoded to "GitHub"


inosik:
We host our own GitLab instance and I configured VuePress like this:

```js
module.exports = {
  themeConfig: {
    repo: 'https://gitlab.my-company.local/my-team/our-product'
  }
}
```

The link to the repository still says "GitHub", because [it's hardcoded in `NavLinks.vue`][gh-link].

  [gh-link]: https://github.com/vuejs/vuepress/blob/c83b95173ce0361ba0b964c67d0c6181f7dffd54/lib/default-theme/NavLinks.vue#L17

ycmjason:
Fixing this at https://github.com/vuejs/vuepress/pull/168

PR #168, opened 2018-04-20
Changing repo label depending on the host

ycmjason:
Fixing #167

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 .temp
 vuepress
 TODOs.md
+*.sw*
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ node_modules
 .temp
 vuepress
 TODOs.md
-*.sw*
--- a/lib/default-theme/NavLinks.vue
+++ b/lib/default-theme/NavLinks.vue
@@ -108,12 +108,12 @@ export default {
     display inline-block
     margin-left 1.5rem
     line-height 2rem
-  .github-link
+  .repo-link
     margin-left 1.5rem
 
 @media (max-width: $MQMobile)
   .nav-links
-    .nav-item, .github-link
+    .nav-item, .repo-link
       margin-left 0
 
 @media (min-width: $MQMobile)
--- a/lib/default-theme/NavLinks.vue
+++ b/lib/default-theme/NavLinks.vue
@@ -78,8 +78,10 @@ export default {
       }
     },
     repoLabel () {
-      if (!this.repoLink) return;
-      if (this.$site.themeConfig.repoLabel) return this.$site.themeConfig.repoLabel
+      if (!this.repoLink) return
+      if (this.$site.themeConfig.repoLabel) {
+        return this.$site.themeConfig.repoLabel
+      }
 
       const repoHost = this.repoLink.match(/^https?:\/\/[^/]+/)[0]
       const platforms = ['GitHub', 'GitLab', 'Bitbucket']
--- a/lib/default-theme/NavLinks.vue
+++ b/lib/default-theme/NavLinks.vue
@@ -78,6 +78,8 @@ export default {
       }
     },
     repoLabel () {
+      if (this.$site.themeConfig.repoLabel) return this.$site.themeConfig.repoLabel
+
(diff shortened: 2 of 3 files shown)


inosik:
Maybe this should be configurable?

ycmjason:
maybe I will keep the logic now since it is quite nice to have less config. But with the config would be a fallback? would that make sense?

inosik:
I'd just keep it simple and let the user configure it with a `repoLabel` option. But maybe it's just me 😄

ycmjason:
I personally prefer just setting the`repoLink` then everything else works. Instead of having needed to set the `repoLabel`. Let me just quickly add this. We can wait for more opinions.

lib/default-theme/NavLinks.vue
```diff
@@ -69,14 +69,22 @@ export default {
...
           : `https://github.com/${repo}`
       }
-    }
+    },
+    repoLabel () {
+      if (this.$site.themeConfig.repoLabel) return this.$site.themeConfig.repoLabel
+
+      const repoHost = this.repoLink.match(/^https?:\/\/[^/]+/)[0] || 'github'
+      return ['GitHub', 'GitLab', 'Bitbucket'].find(platform => {
+        return repoHost.toLowerCase().includes(platform.toLowerCase())
```
ulivz: In client side you shouldn't use `includes` since Buble doesn't ployfill for that. 

BTW, why define a Uppercase constant and then `toLowerCase`?

lib/default-theme/NavLinks.vue
```diff
@@ -69,14 +69,22 @@ export default {
...
         return /^https?:/.test(repo)
           ? repo
           : `https://github.com/${repo}`
       }
-    }
+    },
+    repoLabel () {
+      if (this.$site.themeConfig.repoLabel) return this.$site.themeConfig.repoLabel
+
+      const repoHost = this.repoLink.match(/^https?:\/\/[^/]+/)[0] || 'github'
```
ulivz: Since repo link start with `https://`, so the check `match(/^https?:\/\/[^/]+/)[0]` is duplicated. 

in other words, the statements `match(/^https?:\/\/[^/]+/)[0]` is dangerous since `match` would return undefined!

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules
 .temp
 vuepress
 TODOs.md
+*.sw*
```
ulivz: Why adding this?

ycmjason: For vim swap files.

lib/default-theme/NavLinks.vue
ycmjason: Oh okay, I would change this to `match` instead.

Defining uppercase constants since they are what we want to show on the header, lowercase since we want to compare. But actually using regex with case insensitive flag and match would be nicer.

lib/default-theme/NavLinks.vue
ycmjason: This is true. Will fix this later.

.gitignore
ulivz: You should add it at your project instead of here.

ycmjason: Just thought it would be great since there might be other contributors who use vim.

ulivz: Hmmm, so should I put all the different needs here?

ulivz: It doesn't depend on me. please do not kidnap thinking.

Just I believe most VIM programmers have set this to a global configuration instead of duplicate it at every project.

ycmjason: That is certainly not true for me. I am happy to remove that if this is not wanted.

ycmjason:
@ulivz I have made some changes. Please review them again.

Here are some changes:
- Added a guarded clause to avoid `this.repoLink.match` from failing since we could be confident that `this.repoLink` will always start with `https://`.
- Used for-loop to achieve the same functionality of `Array.prototype.find`.
- Used `RegExp.prototype.test` to check if the host of the `repoLink` contains any recognised platform, if not return `"Source"`.

ulivz:
Hmmmm....  this commit introduced a very obvious bug:

image

Fixed at f55fa00f5d315ed71d39a476078d56938102e131.

PR #180, opened 2018-04-20
fix: prioritize frontmatter's title and lang (close: #177)

### LINKED ISSUE #177: How to change the lang attribute on <html>
I've tried to change the `lang` attribute on `<html>` using YAML front matter, but I probably miss something because it doesn't work.

This is what I do for each page:

```
---
lang: fr-FR
---
```

`$page.frontmatter.lang` is then equal to `fr-FR`, but `$lang` is still `en-US`.

The only way I've found is to use the `locales` option in `.vuepress/config.js`:
```
locales: {
  "/": {
    lang: "fr-FR"
  }
}
```

but it's weird, and activate the `languages` link in the navbar.

ulivz:
Fixed #177

### DIFF
--- a/lib/app/dataMixin.js
+++ b/lib/app/dataMixin.js
@@ -47,13 +47,13 @@ export default {
       return targetLang || defaultLang || {}
     },
     $title () {
-      return this.$localeConfig.title || this.$site.title || ''
+      return this.$page.frontmatter.title || this.$localeConfig.title || this.$site.title || ''
     },
     $description () {
-      return this.$localeConfig.description || this.$site.description || ''
+      return this.$page.frontmatter.title || this.$localeConfig.description || this.$site.description || ''
     },
     $lang () {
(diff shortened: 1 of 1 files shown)


lib/app/dataMixin.js
```diff
@@ -47,13 +47,13 @@ export default {
       return targetLang || defaultLang || {}
     },
     $title () {
-      return this.$localeConfig.title || this.$site.title || ''
+      return this.$page.frontmatter.title || this.$localeConfig.title || this.$site.title || ''
     },
     $description () {
-      return this.$localeConfig.description || this.$site.description || ''
+      return this.$page.frontmatter.title || this.$localeConfig.description || this.$site.description || ''
```
jaylinski: Shouldn't this be `this.$page.frontmatter.description`?

ulivz: Cool, typo. thanks, fixed at 89b8386

PR #185, opened 2018-04-20
fix: Cannot get correct $page when a trailing slash is missed. (close: #183)

### LINKED ISSUE #183: $page variable doesn't gets set probably on urls without a trailing slash
**OS:** Windows 10
**Node:** 8.10.0
**VuePress:**  0.6.1
**Browser:** Chrome 65
Local install
**Package manager:** Yarn

**Bug:**
If you visit a route without a trailing slash, the $page variable doesn't gets set probably and the page won't render correctly.
You also don't get redirected to the path with the trailing slash.

**How to reproduce:**
1.  Clone offical repository
2.  Run dev server (yarn dev)
3.  Open a route without the taling slash (e.g. localhost:8080/guide )

**Expected Behavior:**
You get redirected to the correct path (with the trailing).
This would fix potential SEO problems as well (for more infos: https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html?m=1 )

Issue is linked to: #144

ulivz:
Fixed #183

### DIFF
--- a/lib/app/util.js
+++ b/lib/app/util.js
@@ -1,3 +1,9 @@
+export function ensureEndingSlash (path) {
+  return /(\.html|\/)$/.test(path)
+    ? path
+    : path + '/'
+}
+
 export function pathToComponentName (path) {
   if (path.charAt(path.length - 1) === '/') {
     return `page${path.replace(/\//g, '-') + 'index'}`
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks, I think it's a more complete fix to always ensure the slash at the end (see 99bc0aa)

PR #188, opened 2018-04-20
docs: clarify documentation on how to use links

bencodezen:
When I was first getting setup with Vuepress, I found it a little confusing on how to link out to the relative markdown files in the directory. It took some exploration and experimentation to figure out what exactly worked, so I wanted to provide an easy to follow explanation to reduce that friction for future users.


yyx990803:
👍

PR #155, opened 2018-04-19
feat: Add docsRepo

wangyi7099:
In some cases, project and doc are not in the same repo like Vue. So, and a `docsRepo` option  in config file to distinguish project and doc address.

PR #191, opened 2018-04-21
Allow to use a different docs repo for edit-link

egoist:


egoist:
Hmm didn't notice #155

PR #199, opened 2018-04-21
fix the "github link" on navbar not right showing on mobile device

wicowen:
the github link not get padding class

screen shot 2018-04-22 at 1 43 00 am

after add github-link cssl class

screen shot 2018-04-22 at 1 43 35 am


ulivz:
Thanks, it has been fixed at: f55fa00f5d315ed71d39a476078d56938102e131

PR #213, opened 2018-04-22
Added deploy guide for firebase

ycmjason:


mdaffin:
Should the `.firebaserc` and `firebase.json` files not exist at the project root level? With `"public": "docs/.vuepress/dist",` (or where your actual dist exists) in the firebase.json? Then there is no need for a wrapper script and `firebase deply` can be added to the `package.json`'s `script` section directly.

ycmjason:
Good idea! Will change this now! 🥂

mdaffin:
Much better, though I think you should use `docs/.vuepress/dist` to keep it consistent with the netlify instructions.

ycmjason:
@mdaffin 
Yeah you are right. I have also fixed github pages script for not using `docs/.vuepress/dist`.

PR #215, opened 2018-04-23
Update isActive logic to always checking equality

onevcat:
Current `isActive` checking will fail when a path is prefixed by another. It causes a problem when showing active indicator in the side bar.

eg:

Say we have two pages under `api` folder, which are `uniwebview.md` and `uniwebviewmessage.md`. When accessing `api/uniwebviewmessage.html`, the result when determining side bar active state for `uniwebview` page is:

routePath: "/api/uniwebviewmessage"
pagePath: "/api/uniwebview"

This leads to both uniwebviewmessage and uniwebview be rendered as actived, which is not correct:

snip20180423_1

By changing it to `===`, it goes fine. However, I am not sure whether there would be a regression on this change or not, since there seems like lack of tests for this project.

### DIFF
--- a/lib/default-theme/util.js
+++ b/lib/default-theme/util.js
@@ -46,11 +46,7 @@ export function isActive (route, path) {
   }
   const routePath = normalize(route.path)
   const pagePath = normalize(path)
-  if (endingSlashRE.test(routePath) || endingSlashRE.test(pagePath)) {
-    return routePath === pagePath
-  } else {
-    return routePath.indexOf(pagePath) === 0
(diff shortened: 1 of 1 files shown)


ycmjason:
Not so sure why originally it checks equality only when either of them ends with `/` and check prefix otherwise. But your code looks alright to me, I have tested with my own site and it works ok.

yyx990803:
Thanks, originally this function was used for `isActive` check for the header nav links as well, so it has as some left-over logic 😅

Issue #198, opened 2018-04-21
[Discussion] Roadmap


ulivz:
Hey guys, about the next direction or planning of VuePress, specially set up a **"channel"** to discuss here.

Above all, we need to make clear that the original intention of VuePress is to write documents to Vue and its sub projects. As mentioned in the document, VuePress is composed of two parts: 

1. A static site generator
2. A vue-styled default theme

If we leave everything to the default theme, then when you decide to develop a new theme, you would repeat many things. so for some of our new features, we should consider whether it is handled in `1` or `2`. 

In fact, I expect to complete the development of the core functionality as soon as possible, Only core is STABLE can we focus on **migrating the documents** and **blog theme development** (Maybe need to create a separate repo called `vuepress-theme-blog`). Of course, the premise is that we have prepared useful enough data and APIs for the beginning of the custom theme. 

At present, the core function I think is probably only **Algolia DocSearch Integration**. If you think there are other deserved suitable core needs, feel free to tell us here.

Regards,
Thanks.

## Radmap

- 0.8.0
    - Algolia DocSearch Integration.

dgpgdev:
nice to centralize info :) .

I'll just finish to create a vue cli plugin for vuepress maybe add that inside project features.  https://github.com/dgpgdev/vue-cli-plugin-vuepress

dgpgdev:
Before work algolia docsearch maybe merge pullrequest validated. I don't think blog are useless... vuepress is perfect for documentation.

yyx990803:
One extra thing we need to finalize in core: API for custom themes to write additional files during build.

mdaffin:
> so for some of our new features, we should consider whether it is handled in 1 or 2.

Or we can implement things in a theme, then once they are proven or popular look to migrating them into the core. As long as the core is flexible enough for themes to do this extra useful common features can be backported into the core so that more themes can take advantage of them.

For these, we need a way for themes to write additional files like @yyx990803 suggests. I have attempted to introduce this in #196 but we should probably take about the implementations details a bit more. In short, it loads `themeDir/index.js` and `.vuepress/index.js` and expects them to export a function that takes `{options}` as an argument. This function runs in the context of the server as part of the prepare method so has access to write files to the `outDir` and otherwise manipulate the set options before the site is generated (allowing injection of extra headers or other things).

One thing it does not expose is a way to hint to the build about extra routes that should be generated (such as routes added in enchanceApp.js like in #143)

ulivz:
[...]

PR #216, opened 2018-04-23
Add external plugin config

dacsang97:
I added external plugin config for Vuepress base on `options` variable.
Plugin system can use the local plugin or plugin package in NPM.
**.vuepress/config.js**
``` js
module.exports = {
  ...
  plugins: [
    // Use local plugin. Example: `localPlugin.js` in `.vuepress`
    './localPlugin.js',
    {
      // Use plugin from npm.
      resolve: 'vuepress-plugin-rss',
      // Plugin option
      options: {
        base_url: '/', // required
        site_url: 'https://coralo.io', // required
        copyright: '2018 Coralo' // optional
      }
    }
  ]
}
```
I just public [vuepress-plugin-rss](https://github.com/dacsang97/vuepress-plugin-rss) package for generating `rss.xml` for site using Vuepress site.


mdaffin:
Similar to and will probably conflict with #196 - we really need to discuss how we want plugin support to work in vuepress. You should add your thoughts to #198.

ycmjason:
I think regarding plugins, it would be really cool if vuepress could detect plugins itself. Similar to how parcel does it. Maybe as soon as a `vuepress-plugin-*` is installed in package.json, then load it automatically.

dacsang97:
Oh, I had seen your work, this is an awesome idea. I think my pull request as the extension of your work, because of I config for both local and remote plugin, I think we can public our plugin for the community and reuse them, instead of just using local plugin config.

Issue #222, opened 2018-04-23
Markdown component 


ycmjason:
Hello, 

I think it would be awesome if we could have a markdown vue component that could be used in the markdown pages. 

This way we could module our very long markdown into smaller sections.

Thoughts?

E.g.

```
# My very long md

This is a freaking long md...

<vp-md src="./sectionA.md"/>

<vp-md src="./sectionB.md"/>

<vp-md src="./sectionC.md"/>
```

yyx990803:
I don't think this has to be a component, can simply be a comment-based directive:

``` md
<!-- include ./sectionA.md -->
```

And then parsed and handled directly inside `lib/webpack/markdownLoader.js`.

ycmjason:
I will look into that tonight if I got the time.

ycmjason:
An attempt in here https://github.com/vuejs/vuepress/pull/227.

MartinMuzatko:
There is already https://www.npmjs.com/package/markdown-it-include
But I see a few problems with this:

1. no vue templating within those included markdown files.
2. partials have to be excluded from getting rendered by vuepress

ycmjason:
@MartinMuzatko my attempt at #227 should be able to load vue templating.

Yet, the partials are also being rendered. Maybe we could make it so that vuepress will ignore `.partial.md` files and not render them.

eyleron:
How about the convention in other template languages or SCSS where filenames with preceding underscore don't get rendered?
``` _layout-header.md
```

ycmjason:
@eyleron I am not so sure about this as @yyx990803 and @ulivz haven't got time to advice on this.

davemacdo:
I know this is an old Issue, but it's something I would definitely use. I did want to point out that there are already several syntaxes for this in various markdown flavors. Personally, I'm used to using Mulitmarkdown's [file transclusion syntax](http://fletcher.github.io/MultiMarkdown-5/transclusion.html) in a lot of my markdown already: 

```
# Parent Markdown Stuff

{{path/to/included.md}}
```

See also: 

- Brett Terpstra's [multi-file document syntax](http://marked2app.com/help/Multi-File_Documents.html), which has been adopted by other markdown processors
- iA Writer's [Markdown Content Blocks](https://github.com/iainc/Markdown-Content-Blocks)

A lot of the content I'm including in VuePress sites exists as Markdown already, so the more it can mimic the behavior of other renderers, the better.

ycmjason:
[...]

PR #227, opened 2018-04-23
Attempt to enable inclusion of partial markdown

ycmjason:
> Very unstable, need more advice

I have attempted to add this functionality as discussed in https://github.com/vuejs/vuepress/issues/222.

Current implementation facilitates the below functionality:
1. Replaces `<!-- include ./path/to/my/file.md -->` with the content of the file (recursively)
2. Hot module reloading of dependencies

There are currently 2 known problems in this implementation:
1. Do not utilise the cache properly, files are being read multiple times even when no content is changed. (Slow down the build process.)
    - Maybe I could create another cache in `mdExplodeIncludes()`
2. `<!-- include PATH -->` could not be escaped even in code block. If `PATH` is an actual file, the file content would be used. This behaviour might not be desirable in some occasion. (e.g. when I was documenting about this functionality.)
    - Not sure how should I determine when to escape...


lib/webpack/util.js
```diff
@@ -0,0 +1,27 @@
...
+const { resolve, dirname } = require('path')
+
+const mdExplodeIncludes = exports.mdExplodeIncludes = ({ cwd, src }) => {
+  const deps = []
+
+  return {
+    explodedSrc: src.replace(/<!-- *include +(.*?) *-->/g, (match, path) => {
+      try {
+        const absolutePath = resolve(cwd, path)
+        const content = readFileSync(absolutePath, 'utf8')
```
ulivz: Using a unified convention at the same project. either that all are `await readFile` or all are `readFileSync `

lib/webpack/markdownLoader.js
```diff
@@ -5,26 +5,32 @@ const { EventEmitter } = require('events')
...
 const { inferTitle, extractHeaders, parseFrontmatter } = require('../util')
 const LRU = require('lru-cache')
+const { mdExplodeIncludes } = require('./util')
 
 const cache = LRU({ max: 1000 })
 const devCache = LRU({ max: 1000 })
 
 module.exports = function (src) {
+  const file = this.resourcePath
+  const cwd = path.dirname(file)
```
ulivz: just leverage `dir`

lib/webpack/util.js
```diff
@@ -0,0 +1,27 @@
+const { readFileSync } = require('fs')
+const { resolve, dirname } = require('path')
+
+const mdExplodeIncludes = exports.mdExplodeIncludes = ({ cwd, src }) => {
+  const deps = []
+
+  return {
+    explodedSrc: src.replace(/<!-- *include +(.*?) *-->/g, (match, path) => {
```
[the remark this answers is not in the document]
ycmjason: Good idea. :tada:

lib/webpack/util.js
```diff
@@ -0,0 +1,49 @@
+const { promisify } = require('util')
```
yyx990803: You can use `fs-extra` here without having to `promisify`.

lib/webpack/util.js
ycmjason: I have made this loader asynchronous. And it took me a really really long time to complete the `asyncReplace` function. I am so glad that I did this in such an elegant way. 

Also it took me really long to debug the async loader as I forgot to call the callback when the loader hit the cache. :sob: 

It is using readFile now.

lib/webpack/markdownLoader.js
ycmjason: Can you explain more on this?

lib/webpack/util.js
ycmjason: Oh, didn't know it exists. I will do this now.

ycmjason:
@yyx990803 @ulivz 
I would need some advice from you guys as mentioned in the main thread. :grinning: 

Do you guys think those are big concerns?

Furthermore, one more thing to consider, should we force every partial to end with `.partial.md` and avoid vuepress from rendering a page for it? As discussed in https://github.com/vuejs/vuepress/issues/222.

lib/webpack/markdownLoader.js
```diff
@@ -5,26 +5,35 @@ const { EventEmitter } = require('events')
...
+const { mdExplodeIncludes } = require('./util')
 
 const cache = LRU({ max: 1000 })
 const devCache = LRU({ max: 1000 })
 
-module.exports = function (src) {
+module.exports = async function (src) {
+  const cb = this.async()
+  try {
+  /* IMPORTANT: I didn't indent these lines to hopefully get a better looking diff */
```
ycmjason: I have un-indented the lines in the `try { ... } catch` block to show a better looking diff. Please re-indent when PR is accepted. :blush:

ycmjason:
Regarding the caching problem, after some think about it, it is probably not a problem because a partial is expected to be part of a longer markdown, therefore each partial should be included only once. :taco:

ycmjason:
@kevinmpowell 
You can find the roadmap here. https://github.com/vuejs/vuepress/issues/421

Hopefully this will get released in 1.0.

lib/webpack/markdownLoader.js
```diff
@@ -5,26 +5,36 @@ const { EventEmitter } = require('events')
...
+const { mdExplodeIncludes } = require('./util')
 
 const cache = LRU({ max: 1000 })
 const devCache = LRU({ max: 1000 })
 
-module.exports = function (src) {
+module.exports = async function (src) {
+  const cb = this.async()
+  try {
+/* IMPORTANT: I didn't indent these lines to hopefully get a better looking diff */
```
ycmjason: *Please re-indent these lines before merge!*

ndyakov:
@ycmjason Is there a circular dependency check? I failed to see one.

neumayr:
Maybe the 'Import Code Snippets' feature can be extended to support `md`?

> You can import code snippets from existing files via following syntax:
> `<<< @/filepath`

📝 https://vuepress.vuejs.org/guide/markdown.html#import-code-snippets

ycmjason:
@neumayr I think that's a great idea!

ndyakov:
@neumayr , @ycmjason If I import code I would expect it to be `code` (i.e. shown in a <pre> and highlighted), so if I import `md` I would expect it to be raw (e.g. a doc about how to use markdown imports `md` snippet). `<!-- include... -->` makes more sense for `md` partials.

davemacdo:
Sorry I'm late to the party, but I wanted to point out that there are several other markdown file inclusion syntaxes that exist already. Lots of my VuePress content files are existing markdown documents, and I wonder if it might make the content more "portable" if the implementation used an existing syntax. Multimarkdown uses `{{ file/path.md }}`, which might be a little confusing in a Vue context. There is also Brett Terpstra's [multi-file document syntax](http://marked2app.com/help/Multi-File_Documents.html) and iA Writer's [Content Block syntax](https://github.com/iainc/Markdown-Content-Blocks).

MartinMuzatko:
I agree @davemacdo 
Potentially, this could also be a function called via Vue template binding: {{ $markdown('partial.md') }}
No need to introduce special syntax.

ycmjason:
@davemacdo @MartinMuzatko 
Hello, I just came back from holidays. And I think you made a legit point. I personally quite like the syntax:

```
<<[path/to/file.md]
```

But I guess I will not make the changes to this PR unless @ulivz agree to it. 👍

dovy:
Any progress on this?

timaschew:
@ycmjason

Since there is a similar syntax already to [include code snippets](https://v1.vuepress.vuejs.org/guide/markdown.html#import-code-snippets) which is

```
<<< @/filepath
```

it would be nice to have something similar to include markdown files.

But since v1 is not stable yet this can also be changed later (in another PR), right?

In general I like the approach of using a syntax which works when the markdown is rendered also on GitHub. It allows really easy contribution. Here is an example:

1. Open this page: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/tutorial-manage-vm
2. Click on the Edit button 
3. Spot the `[!INCLUDE cloud-shell-try-it.md]` in the rendered markdown. You can click on it and you'll be able to edit that file.

The syntax they use is: 
```
[!INCLUDE [cloud-shell-try-it.md](../../../includes/cloud-shell-try-it.md)]
```

related: https://github.com/vuejs/vuepress/issues/1474

timaschew:
Does the current implementation allow to include markdown files which itself have references to other resources, like links to other markdown files and images (or even nested includes)?

aliniacb:
We can use the markdown-it-include plugin to achieve this. The only issue is the rebuild process. I have opened an issue  [here](https://github.com/vuejs/vuepress/issues/1616)

Issue #187, opened 2018-04-20
Add navbar disabling option


mvaello:
In the same way as sidebar 'sidebar: false', IMHO, It would be nice to have an option for disable top navbar.

PR #232, opened 2018-04-23
Navbar disabling option (#187)

patzick:
My first contrib to the Vue community ;)
Picked issue with `contribution welcome` label - #187


patzick:
@ycmjason added in new PR, thanks for suggestion!

PR #235, opened 2018-04-24
Change theme enhance app file name to index.js

dacsang97:
Change `enhanceApp.js` to `index.js` for **Theme Level Enhancements** in order to sync with documentation
> Themes can extend the Vue app that VuePress uses by exposing an `index.js` file at the root of the theme.


ulivz:
Thanks, but we have decided that`index.js` will be used as the entry of a theme's plugin.

PR #240, opened 2018-04-24
Lifecycle-based Plugin Support 

ulivz:
# Background

We have discussed a lot about the **plugin**:  [the long discussion & vote](https://github.com/vuejs/vuepress/issues/198#issuecomment-383537304), and these awesome PRs:  #196  and #216, 

But when I start to develop plugin at many different situations, I find that the really appropriate plugin mechanism must delay the timing of execution. i.e. we need to support `life cycle` to let plugin developers to be able to do what they want to do at the right moment.

In order to better describe the PR, I drew these two pictures.

<br>

<p align=center><b>vuepress dev &nbsp;&nbsp;&nbsp;<b/></p>
<p align=center></p>

<br>
<p align=center><b>vuepress build &nbsp;&nbsp;&nbsp;<b/></p>
<p align=center></p>

As shown above, and as we discussed earlier, if we just execute all the plugins at the end of the `prepare`, then once a plugin has written some files to outDir, then you will never see it, which is too bad.

So for now, a plugin would be like this:

```js
module.exports = async function vuepressPluginCustomDomain({ domain } /** only extra plugin options**/) {
  this

   // Execute when prepare is finished, if it's in `build`, the ourDir has been emptied.
   // You do mix extra data and options here before APP launched.
  .ready(() => {
    console.log('Used vuepress-plugin-custom-domain')
  })

  // Execute at the first end of webpack compilation.
  // You can generate extra files here.
  .compiled(async ({ outDir }) => {
    if (process.env.NODE_ENV !== 'production') {
      return
    }
    await fs.ensureDir(outDir)
    await fs.writeFile(path.resolve(outDir, 'CHAME'), domain, 'utf-8')
    console.log('Generated CHAME file.')
  })

   // Execute at each update, only for `dev` 
  .updated(() => {
    console.log('Updated')
  })
}
```

What are the benefits of doing so? for example, if we make the internal log at a default hook handler when we detect there is no other hook handlers, you can override it via plugins, so there would born a separate very handsome plugin.

Of course, It's still worth discussing how to make it better, feel free to give your thoughts.

Thanks~

### DIFF
--- a/docs/.vuepress/vuepress-plugin-custom-domain.js
+++ b/docs/.vuepress/vuepress-plugin-custom-domain.js
@@ -1,8 +1,23 @@
 const path = require('path')
-const fs = require('fs')
+const fs = require('fs-extra')
 
-module.exports = async function vuepressPluginCustomDomain({ outDir, pluginOptions: { domain } }) {
-  if (process.env.NODE_ENV === 'production') {
+module.exports = async function vuepressPluginCustomDomain({ domain }) {
+  this
+
+  .ready(() => {
+    console.log('Used vuepress-plugin-custom-domain')
+  })
+
+  .compiled(async ({ outDir }) => {
+    if (process.env.NODE_ENV !== 'production') {
+      return
+    }
+    await fs.ensureDir(outDir)
     await fs.writeFile(path.resolve(outDir, 'CHAME'), domain, 'utf-8')
-  }
+    console.log('Generated CHAME file.')
+  })
+
+  .updated(() => {
+    console.log('Updated')
+  })
 }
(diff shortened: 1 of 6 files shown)


mdaffin:
Nice work.

One missing bit is to decide if we want to load the theme as a plugin (if say `index.js` exists in the theme root) or otherwise allow themes to add to the list of plugins to load. But this could be done as a future pull request.

docs/.vuepress/vuepress-plugin-custom-domain.js
```diff
@@ -0,0 +1,23 @@
...
+  .ready(() => {
+    console.log('Used vuepress-plugin-custom-domain')
+  })
+
+  .compiled(async ({ outDir }) => {
+    if (process.env.NODE_ENV !== 'production') {
+      return
+    }
+    await fs.ensureDir(outDir)
+    await fs.writeFile(path.resolve(outDir, 'CHAME'), domain, 'utf-8')
```
meteorlxy: `'CNAME'` maybe

PR #241, opened 2018-04-24
fix typo

tianyong90:
I guess this is a small mistake caused by `copy & paste` operate. 😄


yyx990803:
This is intentional because `DANGER` looks a bit weird... it's actually just a more severe type of warning. You can overwrite the label anyway.

PR #243, opened 2018-04-24
Add number in Steps home

luisDanielRoviraContreras:


yyx990803:
Thanks for the PR but I don't think we need to take it that literally...

PR #247, opened 2018-04-25
add cursor pointer in li

luisDanielRoviraContreras:
Only the element `a` has the cursor pointer and it seems that you can only click on it and it can be in all the `li`


jaylinski:
It would be better to make the a-tag use the full width.

PR #253, opened 2018-04-25
docs: french documentation

### LINKED ISSUE #252: docs: add french documentation
Provide french documentation.

Currently there's `english` and `chinese`

lbenie:
closes #252


ulivz:
Original words at [#153 (korean translation)](https://github.com/vuejs/vuepress/pull/153#issuecomment-382979879) from Evan:

> Thanks for the contribution. Since VuePress is still in early iterations, our documentation may change a lot, and making sure translations stay updated can be quite difficult, especially when I cannot understand the translation.
So for now, we are only keeping the Chinese translation because I am able to update it along with the English version when we make changes to the API. We won't be accepting more translations until we are stable. Translations for other languages can be done in forks for now.

So you can continue to work for it at your fork branch, we will recored each awesome contributor who are working for the translation. we will notify you when VuePress is stable.

Thanks again for great work!

Regards,
VuePress Team.

PR #254, opened 2018-04-25
tweak: make the SearchBox of default theme better in narrow screen

meteorlxy:
- cursor on search box should be pointer when it is not focused in narrow screen (only the icon is displayed)
- when the screen width is between `$MQNarrow` and `$MQMobile`, the `.suggestions` box may overflow the left edge


ulivz:
Thanks for this work, but it would be better to provide the contrastive snapshots when did any UI change:

# Before

image

# After

image

meteorlxy:
Got it~ Thanks~

Issue #238, opened 2018-04-24
Sidebar header extraction improvements


yyx990803:
Currently with the following titles:

``` md
## `some code`

## &lt;foo&gt;

## [Some link](/)
```

When extracted into the sidebar they are displayed verbatim. We need to improve the title inference / header extraction logic to handle these cases (Emoji was fixed previously in https://github.com/vuejs/vuepress/pull/206 but it was only one of the many possible cases).

jlooper:
What if one could write something like this:

## `some code` | This is wicked cool code

Then the part after the pipe could be the title? I've been tinkering with side bars recently...

ulivz:
Cool, some questions:

1. Just removing `, or keep the `<code>` style?

```
## `some code`
```

2. Just unescaping it ?

```
 ## &lt;foo&gt;
```

3. Just handled as a nav link ?

```
## [Some link](/)
```

PR #271, opened 2018-04-26
feat: header extraction improvement (close: #238)

ulivz:
# Summary

Fixed: #238 
Test Header:

```md
## `How` [It](/app/) &lt;Works&gt; **app** _italic_ *italic*
```

# Snapshot

### Before

image

### After

image

### DIFF
--- a/lib/util/parseHeaders.js
+++ b/lib/util/parseHeaders.js
@@ -19,8 +19,8 @@ const removeMarkdownToken = str => str
 
 // put here to avoid circular references
 const compose = (...processors) => {
-  if (processors.length === 0) input => input
-  if (processors.length === 1) processors[0]
+  if (processors.length === 0) return input => input
+  if (processors.length === 1) return processors[0]
   return processors.reduce((prev, next) => {
     return (...args) => next(prev(...args))
   })
--- a/lib/util/parseHeaders.js
+++ b/lib/util/parseHeaders.js
@@ -11,8 +11,11 @@ const unescapeHtml = html => html
   .replace(/&gt;/g, '>')
 
 const removeMarkdownToken = str => str
-  .replace(/`(.*)`/, '$1')
-  .replace(/\[(.*)\]\(.*\)/, '$1')
+  .replace(/`(.*)`/, '$1')  // ``
+  .replace(/\[(.*)\]\(.*\)/, '$1')  // []()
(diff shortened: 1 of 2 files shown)


lib/util/parseHeaders.js
```diff
@@ -0,0 +1,29 @@
...
+
+const removeMarkdownToken = str => str
+  .replace(/`(.*)`/, '$1')  // ``
+  .replace(/\[(.*)\]\(.*\)/, '$1')  // []()
+  .replace(/\*\*(.*)\*\*/, '$1')  // **
+  .replace(/\*(.*)\*/, '$1')  // *
+  .replace(/_(.*)_/, '$1')  // _
+
+// put here to avoid circular references
+const compose = (...processors) => {
```
ycmjason: The checking of the `processors.length` seems unnecessary. Perhaps something like this?

```js
const compose = (...processors) => {
  return processors.reduce((prev, next) => {
    return (...args) => next(prev(...args))
  }, x => x)
}
```

Or

```js
const compose = (...processors) => {
  return [
    x => x,
    ...processors
  ].reduce((prev, next) => {
    return (...args) => next(prev(...args))
  })
}
```

ulivz: Hmmm? Are you telling me how to write code? 

Code should be as easy as possible to understand, don't always think about drag racing.

If you expect me to make a change, first give redux a PR.

yyx990803: @ulivz I also prefer easier to write code, but this is open source and @ycmjason is just offering what he thinks might be better code. He is not telling you how to write code and you don't have to be combative about this.

ulivz: @yyx990803 Yes, I should say convincing reasons instead of arguing for it.  

@ycmjason I'm terribly sorry about my recklessness, and I liked the style what you said above, but, after a long time, I found I always need to think back to what parameters of reduce's callback needs to accept. Until one day I read the source code of Redux, I found it's easier to write. so I'd like this style.

lib/util/parseHeaders.js
```diff
@@ -0,0 +1,29 @@
...
+  if (processors.length === 0) input => input
```
ycmjason: Just curious, did you miss `return` here? And the next line?

ulivz: Nice! Thanks! fixed at a8c6959

PR #272, opened 2018-04-27
Active sidebar link highlight

ryanriv:
This is still a work in progress, but I'd really like your feedback on the approach so far.

The main idea here is to gather up all of the anchors, then change the hash when the user scrolls beyond one of the anchor headers. I am filtering the anchors so that we only set the hash to links in the sidebar.

Try it in Chrome/Safari, it should work really nicely! Firefox seems to have different behavior when the hash is changed, though. It jumps to the anchors as soon as they're updated. The result is very jerky scrolling, and scrolling up will immediately skip to the top of a section. Any advice for a better way to do this?

### DIFF
--- a/lib/default-theme/Layout.vue
+++ b/lib/default-theme/Layout.vue
@@ -162,7 +162,10 @@ export default {
       this.setActiveHash()
     }, 300),
     setActiveHash () {
-      const anchors = document.querySelectorAll('.header-anchor')
+      const sidebarLinks = [].slice.call(document.querySelectorAll('.sidebar-link'))
+      const anchors = [].slice.call(document.querySelectorAll('.header-anchor'))
+        .filter(anchor => sidebarLinks.some(sidebarLink => sidebarLink.hash === anchor.hash))
+        
       const scrollTop = Math.max(window.pageYOffset, document.documentElement.scrollTop, document.body.scrollTop)
 
       for (let i = 0; i < anchors.length; i++) {
--- a/lib/default-theme/Layout.vue
+++ b/lib/default-theme/Layout.vue
@@ -29,7 +29,7 @@ import Page from './Page.vue'
 import Sidebar from './Sidebar.vue'
 import { pathToComponentName } from '@app/util'
 import { resolveSidebarItems } from './util'
-import throttle from 'lodash/throttle'
+import throttle from 'lodash.throttle'
 
 export default {
   components: { Home, Page, Sidebar, Navbar },
--- a/lib/default-theme/Layout.vue
+++ b/lib/default-theme/Layout.vue
@@ -162,8 +162,7 @@ export default {
       this.setActiveHash()
     }, 200),
     setActiveHash () {
-      const anchors = gatherHeaderAnchors()
-
+      const anchors = document.querySelectorAll('.header-anchor')
       const scrollTop = Math.max(window.pageYOffset, document.documentElement.scrollTop, document.body.scrollTop)
 
       for (let i = 0; i < anchors.length; i++) {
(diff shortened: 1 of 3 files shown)


lib/default-theme/Layout.vue
```diff
@@ -173,6 +181,35 @@ function updateMetaTags (meta, current) {
...
+
+    if (isActive && window.location.hash !== anchor.hash) {
+      window.location.hash = anchor.hash
+      return
+    }
+  }
+}
+
+function gatherHeaderAnchors () {
+  const sidebarLinks = Array.from(document.querySelectorAll('.sidebar-group-items a.sidebar-link'))
```
ulivz: Use array spread instead of `Array.from`, since buble doesn't polyfill for that in IE11.

lib/default-theme/Layout.vue
```diff
@@ -173,6 +181,35 @@ function updateMetaTags (meta, current) {
...
+
+  const anchors = Array.from(document.querySelectorAll('a.header-anchor'))
+    .filter(x => sidebarLinks.map(x => x.hash).includes(x.hash))
```
ulivz: Same here, `Array.prototype. includes`. you can use `.some` instead

lib/default-theme/Layout.vue
ryanriv: I tried spread initially, but it seems to have some problems with NodeLists (https://gitlab.com/Rich-Harris/buble/issues/81). I would really prefer to use it here instead though. Do you have a preference for something else, assuming spread wouldn't work?

lib/default-theme/Layout.vue
```diff
@@ -29,6 +29,7 @@ import Page from './Page.vue'
 import Sidebar from './Sidebar.vue'
 import { pathToComponentName } from '@app/util'
 import { resolveSidebarItems } from './util'
+import throttle from 'lodash/throttle'
```
ycmjason: Shouldn't this be `lodash.throttle`?

lib/default-theme/Layout.vue
```diff
@@ -111,6 +112,8 @@ export default {
     this.$watch('$page', updateMeta)
     updateMeta()
 
+    window.addEventListener('scroll', this.onScroll)
```
ycmjason: Should this live in `created()` hook? If the user scrolled after `mounted()` but before `created()`, will `this.onScroll` be available?

ulivz: You ask a contributor to access `window ` at `created` hook? Are you kidding me?

meteorlxy: Emmm... In addtion, what is 

> after `mounted()` but before `created()`

ulivz: @ycmjason Pay attention to your wording. If I don't read your comment, then you will be conveying a very low-level error in a vue official project. Is this reasonable? 

BTW, this is not your project, please take a good look at Vue before you speak anything, or I'll think you're just making trouble.

yyx990803: @ulivz please pay attention to yours as well. You are all contributors. You all can make mistakes and can have different opinions. Point out problems with neutral language, don't bring your ego into this.

yyx990803: @ulivz please pay attention to yours as well. You are all contributors. You all can make mistakes and can have different opinions. Point out problems with neutral language, don't bring your ego into this.

ulivz: Thank you. come back and see, I really shouldn't bring my feelings into it, and should be more neutral. I'm sorry.
For above problem, the scroll event should be handled in `mount` or `beforeMount` hook, since we are building a SSR application which has some browser api access restrictions.

ycmjason: Opps, I must have misread previously.

Tried to removed this whole thread but turns out only be able to remove my own comment. I was not asking a contributor to access `window` at `created()`. Just genuinely curious why he put it in `mounted()` as I obviously just got messed up in my head about the order of `created()` and `mounted()` somehow. Thanks @ulivz for clarifying this.

ycmjason: I feel really sorry to make you feel that I am offending you all the time. In case you feel the same again, please do tell me and perhaps suggest a way that I should have put it. 

🎉

ulivz: Work hard together to make this project better. Come on! 🎉

ulivz:
Thanks for the great work! I test it with different browsers in my local area. it looks very good.

PR #282, opened 2018-04-27
feat: generate the timestamp of creation and last updated for each doc file (close #258)

### LINKED ISSUE #258: Feature Request: Optional "last updated" date in default theme
It may be useful to have an optional "last updated" timestamp displayed on a page, especially for docs that may not be tied to specific release versions.

I was thinking this could be automated nicely with git during build on a page-by-page basis with something like:
```bash
git log -1 --format="%ct" [file]
# returns timestamp in seconds of last commit of file
# easily used to create JS Date
```
Could also do the whole docs directory in one batch with something like:
```bash
git ls-tree -r --name-only HEAD docs | while read filename; do echo "$(git log -1 --format="%ct" -- $filename) $filename"; done
# returns lines of "timestamp filepath"
# could be modified to return a JSON array
```
If anyone likes this idea, do you have preferences on where the timestamp would be displayed? Next to the edit link or under page header?

I would be happy to attempt an implementation if this is of interest.

ulivz:
Close #258


meteorlxy:
When a file is saved without any changes, its `mtimeMs` will update, but git won't track the change.

ulivz:
Good point, so there is a discussion which type of time should be provided to user.

yyx990803:
https://github.com/vuejs/vuepress/issues/258#issuecomment-385284682

ulivz:
Updated to git commit time stamp.

inosik:
What if the project uses some other version control system? Or if the project is in its early stages and not under version control yet?

meteorlxy:
@inosik That's it!

ycmjason:
@inosik Interesting thought, I think that's a valid point. Perhaps a fallback to fs would be a nice thing to have.

meteorlxy:
Yeah, I've ever thought about that but forgot to post to that issue.

If a new file is created, but has not been commited into git repo. Then if you are going to `dev` or `build` for test, it may cause some issue.

ycmjason:
@meteorlxy @inosik 
Should we create an issue so that we won't forget about this? 🌵

yyx990803:
Maybe the last updated feature should be off by default, and mention in docs that you need git to use this feature.

ycmjason:
@yyx990803 that would be nice, since fs timestamp is not very informational anyways.

inosik:
> Should we create an issue so that we won't forget about this?

See #340.

ulivz:
@yyx990803 fixed it at #338

PR #283, opened 2018-04-27
Unclear How to Run deploy.sh

yuens1002:
I was unclear on how to run deploy.sh, added a bit of instructions in the doc to clarify the step.


ulivz:
Thanks, but the original version is more easier for us to understand, especially more coherent.

yuens1002:
Run where with what? Is not answered... thus the unclarity.

yuens1002:
Thanks for taking the time to address this concern. Yes yes I agree the doc
needs to be sccuint. However, .sh has never been part of Vue's build
process. Npm is. The doc should mention run (sh filename.sh) in terminal to
deploy. It's a small change that clarifies the last step, at least it is
for me.

Issue #284, opened 2018-04-27
Feature Request: Ability to disable edit links on specific pages.


achendrick:
The majority of the pages on my site will be in a submodule that will be publicly available for editing, while others will be in a private repo that should be maintained privately along with other source.

I'd like to remove the edit links from these private pages so it's not confusing when those pages reach a 404 as they won't exist in the submodule.

ulivz:
Will release it at 0.9.0: #419. 🚀

PR #286, opened 2018-04-27
feat: hide edit link by page (close: #284)

achendrick:
close: #284


lib/default-theme/Page.vue
```diff
@@ -52,14 +52,16 @@ export default {
       }
     },
     editLink () {
+      if(this.$page.frontmatter.editLink === false) {
+        return
+      }
```
achendrick: Good call @ulivz, thanks!

PR #293, opened 2018-04-28
Change title at home

luisDanielRoviraContreras:
Now you can add an html tag in the **heroText** as for example` <b> </ b> `

```yaml
heroText: <b>Vuesax</b> Frontend Vue Components
```

## Before

screenshot_27

## After
screenshot_28

Try another functionality that at least in my case I use a lot that is to add a link

```yaml
heroText: <b>Vuesax</b> Frontend <a target="_blank" href="https://vuejs.org/">Vue</a> Components
```
screenshot_29

### DIFF
--- a/lib/default-theme/Home.vue
+++ b/lib/default-theme/Home.vue
@@ -2,7 +2,8 @@
   <div class="home">
     <div class="hero">
       <img v-if="data.heroImage" :src="$withBase(data.heroImage)" alt="hero">
-      <h1>{{ data.heroText || $title || 'Hello' }}</h1>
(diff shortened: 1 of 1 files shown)


yyx990803:
The YAML is only meant for convenience - you can simply inspect the default generated structure and write your own HTML in the markdown section:

``` html
<div class="hero">
  <img src="/hero.png" alt="hero">
  <h1>VuePress</h1>
  <p class="description">
      Vue-powered Static Site Generator
  </p>
  <p class="action"><a href="/guide/" class="nav-link action-button">Get Started →</a></p>
</div>
```

We will probably add a note about that in the docs.

Issue #101, opened 2018-04-17
Markdown in Frontmatter


MartinMuzatko:
Hey there!

What would you do, if you wanted to highlight a few words in the features of the homepage with markdown? 
Markdown in frontmatter is not parsed and also cannot reliably, possibly. Because some values are turned into the appropriate format. e.g. false -> boolean, or 2018-03-20 is a Date object.
I wonder if I should write custom components and then use them instead of frontmatter, so I can work around the limits like so:

``` html
<Feature>
    <template slot="description">
        **special** features to describe here.
    </template>
</Feature>
```

But then, I could also almost just start writing plain HTML.

I want the flexibility of layouts and metadata attached to articles, but still define and re-use some common components.
Maybe I could also abuse custom containers for this- but they also just define two properties: title and text body. But at least I could leverage markdown for the task.

yyx990803:
Thought a bit about this and I don't think we'll extend the home page functionality any further, because you can just use plain HTML after the frontmatter:

```
---
home: true
---

<div class="features">
  <div class="feature">
    <h2>A feature</h2>
     <p>some <strong>text</strong></p>
  </div>
</div>
```

While being able to write markdown for this is easier, it involves a lot more complexity in the implementation so it's not worth the effort IMO.

MartinMuzatko:
Thanks for the reply @yyx990803, The features were just an example. I agree on not further expanding the homepage.

I use front matter as a mechanism to use the content in different areas of the page. E.g. a description/teaser-text part of the header, and then continue with the usual content.

I was also thinking about implementing this with multiple `<Content />` outlets, like Vues slot mechanism. Maybe that is too complex as a feature. What do you think?

PR #295, opened 2018-04-28
 Be able to use html tags in features, footer

luisDanielRoviraContreras:
Add directive v-html in the features and the footer to use html tags

## Example 

```yaml
- title: <b>Responsive</b> components
  details: We work with Responsive <a target="_blank" href="https://vuejs.org/">Components</a>
```


meteorlxy:
I'm not sure but that could be implemented in custom theme

ulivz:
We should avoid using `v-html`. see https://github.com/vuejs/vuepress/issues/101#issuecomment-382109175.

PR #297, opened 2018-04-29
Update index.js

zuck:
Resolve TypeError if title is not a string.

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -1,6 +1,6 @@
 const parseEmojis = str => {
   const emojiData = require('markdown-it-emoji/lib/data/full.json')
-  return str.replace(/:(.+?):/g, (placeholder, key) => emojiData[key] || placeholder)
+  return String(str).replace(/:(.+?):/g, (placeholder, key) => emojiData[key] || placeholder)
 }
 
 exports.normalizeHeadTag = tag => {


yyx990803:
When does this happen? When submitting PRs, make sure to provide enough context so that we can determine whether this actually fixes the problem.

ulivz:
Please give enough context, thanks!

zuck:
Yeah, sorry guys. I was trying to port an existing (working) project from Hexo to Vuepress.

Following the "getting started" guide I simply ran `vuepress dev` from root folder , receiving the following error:

```
TypeError: str.replace is not a function
    at parseEmojis (/usr/lib/node_modules/vuepress/lib/util/index.js:3:14)
    at exports.inferTitle (/usr/lib/node_modules/vuepress/lib/util/index.js:38:12)
    at Promise.all.options.pageFiles.map (/usr/lib/node_modules/vuepress/lib/prepare.js:193:19)
    at <anonymous>
```

Of course I can guess this is related to the fact the root folder of my existing Hexo project contains many different kind of files, but docs are a little bit vague about project structure for beginners.

meteorlxy:
I find what happen. I have ever met similar problem in another projects.

If you set `title` in the frontmatter of a page (not `home: true`) to a pure `Number` value, it will try to call `.replace()` on a `Number` object, which leads to this error.

ulivz:
@znck Thanks for this PR, but with some conflict and not a complete fix.
@meteorlxy  thanks for your investigation. 

completely fixed at 5023d195dcd7ee6d3710a43558940bcdc31d5a2f

znck:
Not me @ulivz. 

@zuck Thanks for the PR.

PR #294, opened 2018-04-28
Change in description soport html

luisDanielRoviraContreras:
Change to be able to add html tags in the description.

## Example

```yaml
# README.md
tagline: We love what we do. Let us help you do what <b>You love</b>.
```

```js
// .vuepress/config.js
module.exports = {
  description: 'We love what we do. Let us help you do what <b>You love.</b>',
}
```

PR #304, opened 2018-04-30
feat: support custom homepage

ulivz:
# Background 

Yesterday, I thought a log about the `Custom Homepage` before closing a issue (#294). anyhow, we should NEVER use `v-html` to render anything from front-matter —— The YAML is only meant for convenience, and you even cannot use multiple line syntax. 

BTW, Another defect of using html at front-matter is that we will need to parse each field which would be html. in fact, we only need to support HTML for description in homepage. while in other page the description in only existed in `<meta>` tag.

@luisDanielRoviraContreras Thanks for your PR, as Evan said at https://github.com/vuejs/vuepress/pull/293#issuecomment-385284170, we can simply inspect the default generated structure and write your own HTML in the markdown section:

```html
<div class="hero">
  <img src="/hero.png" alt="hero">
  <h1>VuePress</h1>
  <p class="description">
      Vue-powered Static Site Generator
  </p>
  <p class="action"><a href="/guide/" class="nav-link action-button">Get Started →</a></p>
</div>
```

It's a good idea, but I still want to keep some custom functionalities out of the box. which is for a more smooth transition to `pure html` or **Custom Layout**.

So with this PR, you can implement your requirements like this:

```md
---
home: custom
# ... other front-matters
---

<Home>

  <slot slot="title">
    <b>Vuesax</b> Frontend Vue Components
  </slot>
  
  <slot slot="description">
    We love what we do. Let us help you do what <b>You love.</b>
  </slot>

  // ... additional content
  
</Home>
```

> You can use multiple line syntax, router links, Vue components etc.

@yyx990803 This feature only import very small change to the core. and fully downward compatibility. please help me to confirm the document. thanks!

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,5 +1,5 @@
 ---
-home: true
+home: custom
 heroImage: /hero.png
 actionText: Get Started →
 actionLink: /guide/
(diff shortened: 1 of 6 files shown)


ulivz:
Closing this PR since it will make core complex.

Issue #278, opened 2018-04-27
postcss-loader warnings


egoist:
https://github.com/vuejs/vuepress/blob/2d6f02f3b06ffd58d1f0000001a621545229d458/lib/build.js#L99-L107

VuePress didn't handle webpack warnings, but if you add following lines:

```js
if (stats.hasWarnings()) {
  return reject(new Error(stats.toString()))
}
```

And if you run `yarn build --debug` in VuePress repo, there're actually postcss-loader warnings:

<img width="953" alt="2018-04-27 10 04 08" src="https://user-images.githubusercontent.com/8784712/39366600-f755b864-4a66-11e8-8423-3a1e6395c06f.png">

Not sure if this is a vue-loader bug since I can only reproduce it in this repo. I've tried console.log `this.sourceMap` in `vue-loader`, seems it's always `true` even if I already set `webpackConfig.devtool: false`

Test against latest commit: https://github.com/vuejs/vuepress/commit/2d6f02f3b06ffd58d1f0000001a621545229d458

meteorlxy:
And another warning:

`DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead`

ulivz:
Reproduced this problem, It works when set `sourceMap` to `true`. but what is strange is that no source map was generated. maybe need more investigation. 

```js
rule.use('postcss-loader').loader('postcss-loader').options(Object.assign({
        plugins: [require('autoprefixer')],
        sourceMap: true
      }, siteConfig.postcss))
```

meteorlxy:
```js
  if (debug) {
    // config.devtool('source-map')
  } else if (!isProd) {
    config.devtool('cheap-module-eval-source-map')
  }
```

I comment that line and the warning disappear. Maybe the `sourceMap` should be set to `debug`?

ulivz:
Maybe?

```js
  if (!isProd) {
    config.devtool('cheap-module-eval-source-map')
  } else if (debug) {
    config.devtool('source-map')
  }
```

meteorlxy:
emmmm... `dev` command doesn't have `--debug` mode, it only works in `build` command, in which `isProd` is always `true`

meteorlxy:
Oh my god, me neither!!

ulivz:
So amazing, what happened?

meteorlxy:
I have  nooooooooo idea.

ulivz:
Welcome to find it out, thanks!

[...]

PR #305, opened 2018-04-30
fix: postcss-loader warnings about sourceMap (#278)

meteorlxy:
Two main changes:
- fix postcss-loader warnings in build mode
- enable sourceMap in development mode

One thing left:
- How to disable `sourceMap` "indeed" when using `css-loader` with `modules: true`


meteorlxy:
Details explained in #278

ulivz:
Fixed at ac4ce84f38128644345123afaf0955d95093cf30.

Issue #306, opened 2018-04-30
MD files with name ending in `index` fails to build


x8x:
Markdown file which name ends with `index` fails to build:
- `Gitter-Room-Index.md` -> build fail
- `Gitter-Room-Index_.md` -> build ok

```
Rendering page: /data/file.html[vuepress] No matching page found for sidebar item "/data/Gitter-Room-Index.md"
[vuepress] No matching page found for sidebar item "/data/Gitter-Room-Index.md"
[vuepress] No matching page found for sidebar item "/data/Gitter-Room-Index.md"
Error rendering /data/file.html:
TypeError: Cannot read property 'match' of undefined
    at getHash (vuepress/lib/default-theme/util.js:13:0)
    at isActive (vuepress/lib/default-theme/util.js:43:0)
    at render (vuepress/lib/default-theme/SidebarLink.vue?61b3:10:0)
    at renderWithStyleInjection (vuepress/node_modules/.registry.npmjs.org/vue-loader/15.0.4/node_modules/vue-loader/lib/runtime/componentNormalizer.js:78:0)
    at createFunctionalComponent (/vuepress/node_modules/.registry.npmjs.org/vue/2.5.16/node_modules/vue/dist/vue.runtime.common.js:4058:30)
    at createComponent (/vuepress/node_modules/.registry.npmjs.org/vue/2.5.16/node_modules/vue/dist/vue.runtime.common.js:4248:12)
    at _createElement (/vuepress/node_modules/.registry.npmjs.org/vue/2.5.16/node_modules/vue/dist/vue.runtime.common.js:4418:15)
    at createElement (/vuepress/node_modules/.registry.npmjs.org/vue/2.5.16/node_modules/vue/dist/vue.runtime.common.js:4355:10)
    at vm._c (/vuepress/node_modules/.registry.npmjs.org/vue/2.5.16/node_modules/vue/dist/vue.runtime.common.js:4487:42)
    at server-bundle.js:14480:650
    at VueComponent.renderList [as _l] (/vuepress/node_modules/.registry.npmjs.org/vue/2.5.16/node_modules/vue/dist/vue.runtime.common.js:3703:16)
    at VueComponent.SidebarGroupvue_type_template_id_325528d9_render (vuepress/lib/default-theme/SidebarGroup.vue?b7ce:1:547)
[...]

PR #308, opened 2018-04-30
fix: index file judgement bug (#306)

meteorlxy:
#306

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -236,7 +236,7 @@ async function genComponentRegistrationFile ({ sourceDir }) {
   return `import Vue from 'vue'\n` + components.map(genImport).join('\n')
 }
 
-const indexRE = /\b(index|readme)\.md$/i
+const indexRE = /(?<=(^|\/))(index|readme)\.md$/i
 const extRE = /\.(vue|md)$/
 
 function fileToPath (file) {


lib/prepare.js
```diff
@@ -236,7 +236,7 @@ async function genComponentRegistrationFile ({ sourceDir }) {
   return `import Vue from 'vue'\n` + components.map(genImport).join('\n')
 }
 
-const indexRE = /\b(index|readme)\.md$/i
+const indexRE = /(?<=(^|\/))(index|readme)\.md$/i
 const extRE = /\.(vue|md)$/
```
ulivz: `lookbehind` is good, how about `/(^|\/)(index|readme)\.md$/i`? it's same for this case.

meteorlxy: It's not the same. `\b` will not  match the word boundary itself. But `/(^|\/)(index|readme)\.md$/i` will match the begining `/`

meteorlxy: I meant  /(^|\/)(index|readme)\.md$/i too... /(^|\/)(index|readme)\.md$/i has different behavior

meteorlxy: Well, in other words, `\b` and `(?<=(^|\/))` won't include the begining `/`, but `(^|\/)` will

meteorlxy:
Alternative code:
```js
const indexRE = /(^|\/)(index|readme)\.md$/i

// ...

    // README.md -> /
    // foo/README.md -> /foo/
    const fileReplace = file.replace(indexRE, '')
    return '/' + ( fileReplace === '' ? '' : fileReplace + '/')
```

PR #277, opened 2018-04-27
refactor: make the prepare process clearer and modify the resolve alias

meteorlxy:
- Write notfound route to the temp route file instead of adding in `app.js`
- Make the `resolveOptions` process in `prepare` clearer
- Change the `@theme` alias to theme directory. Add `@themeLayout` and `@themeNotFound` aliases for `Layout.vue` and `NotFound.vue` respectively.

PR #313, opened 2018-05-01
Feature: theme-related pages

oott123:
Based on [this](https://github.com/vuejs/vuepress/issues/36#issuecomment-382549812) idea.

Theme can define a list of pages it wants to render, and VuePress will render it during buliding.

I have wrote some docs along with the code in Chinese, please check it out in the code diff or [here](https://deploy-preview-313--vuepress.netlify.com/zh/guide/custom-themes.html#%E6%B8%B2%E6%9F%93%E4%B8%BB%E9%A2%98%E9%A1%B5%E9%9D%A2).


meteorlxy:
I'm waiting for #277  to be merged to add this feature

meteorlxy:
I don't think add the extra pages to `options.siteData.pages` is a good idea. 

For example, if users want to `v-for` all the `pages`, they have to exclude the extra pages manually.

ulivz:
Thanks!

I think we should avoid opening new file-level API for new features. all of them should be transferred to plugin so that we can only focus on designing plugin API.

Issue #160, opened 2018-04-19
Dynamic routes do not get generated with build


mdaffin:
You have access to the route via `.vuepress/enhanceApp.js` and can add extra routes such as:

```
export default ({ Vue, router, options }) => {
  router.addRoutes([
    { path: '/test', redirect: '/' }
  ])
}
```

or 

```
const Test = () => import("./Test");

export default ({ Vue, options, router }) => {
  router.addRoutes([
    { path: "/test", component: Test }
  ]);
};
```

And they work as expected when you run it is dev mode `vurepress dev site`. But when building the app `vurepress build site` the added routes do not get generated/prerendered, resulting in a 404 when trying to load the route for the first time.

As far as I can tell there is no other way to add routes to the generated list of routes which means that routes cannot be added dynamically via the site or themes.

ycmjason:
Maybe try to use the `config.nav` or `config.sidebar`? Not sure if the ssr respects the manual routes from vue router.

mdaffin:
Even it that works there are pages I want to be rendered (redirect links) that I do not want to appear in the sidebar or navy menu. We need a way to add extra routes through the config and via themes for this to work fully.

Bloss:
[util.js](https://github.com/vuejs/vuepress/blob/master/lib/app/app.js)
Here, the router option is been created by `new Router()`, so this router option cannot add extra routes

mdaffin:
@Bloss utils.js or app.js? Not that it matters both of them are client-side scripts and the problem is with the server-side prerendering.

There is no problem with adding new routes in the client-side applications - they work as expected when in dev or SPA mode. The problem is when you go to build the app the server-side it does not know about the extra routes and so does not *prerender* the pages. This results in a 404 when you first hit them (but then work if you navigate to them once in SPA mode).

meteorlxy:
According to the following code, what should be rendered is determined by the `siteData.pages`:

https://github.com/vuejs/vuepress/blob/f0b33e186a062f03469339a4839a1c8c01afc725/lib/build.js#L63-L67

The origin routes file is generated from `siteData.pages`, too:

https://github.com/vuejs/vuepress/blob/f0b33e186a062f03469339a4839a1c8c01afc725/lib/prepare.js#L277-L307

Although your add routes to `router`, the `siteData.pages` won't be aware of that.

meteorlxy:
To make things clear:

Current `routes` flow is -   
(1) source code  
(2) prepare `options`  
[...]

PR #315, opened 2018-05-01
feat: support theme default config and extra pages rendering

meteorlxy:
Three features:
1. Support theme default config by `theme-dir/config.js`
2. Allow custom theme to render extra pages ([#198](https://github.com/vuejs/vuepress/issues/198#issuecomment-383378566)) (Another implementation #313)
3. Allow user to render extra pages based on a theme (#160)

Feature 1 is necessary for Feature 2. I propose to use `theme-dir/config.js`, which is similar to `.vuepress/config.js`, as the theme default config

To achieve Feature 2,  add `extraPages` in `theme-dir/config.js`, and add coresponding extra routes in `theme-dir/enhanceApp.js`. That will enable extra pages rendering for theme.

To achieve Feature 3, add `extraPages` in `.vuepress/config.js`, and add coresponding extra routes in `.vuepress/enhanceApp.js`. That will enable extra pages rendering based on a theme. Notice that theme's `extraPages` config could be overwritten in `.vuepress/config.js`.

E.g: 
```js
// .vuepress/config.js
// or theme-dir/config.js
module.exports = {
  extraPages: [
    '/test'
  ],
  // other config...
}
```

```js
// .vuepress/enhanceApp.js
// or theme-dir/enhanceApp.js
const Test = () => import("./Test");

export default ({ Vue, options, router }) => {
  router.addRoutes([
    { path: "/test", component: Test }
  ])
}
```

Docs will be added after accepted.


meteorlxy:
What do you think about this? @mdaffin

lib/prepare.js
```diff
@@ -335,6 +324,26 @@ function sort (arr) {
...
 
+async function resolveConfig (dir) {
+  const configPath = path.resolve(dir, 'config.js')
+  const configYmlPath = path.resolve(dir, 'config.yml')
+  const configTomlPath = path.resolve(dir, 'config.toml')
+
+  delete require.cache[configPath]
+
+  // resolve siteConfig
+  let config = {}
```
ycmjason: Just my honest opinion 😊❤️🙈, I think 335-344 would be cleaner with the following: 

```js
if (fs.existsSync(configYmlPath)) {
  return await parseConfig(configYmlPath)
}

if (fs.existsSync(configTomlPath)) {
  return await parseConfig(configTomlPath)
}

if (fs.existsSync(configPath)) {
  return require(configPath)
}

return {}
```

meteorlxy: Uh looks better. I just moved things down into a functions without much refactor.

ycmjason: Yea yea! Just spot that when I see the changes, might be worth it to make the change.

meteorlxy:
@ycmjason Maybe we should ensure all the files have been rendered? Or some of the pages may not been rendered before the script exits?

meteorlxy:
@ycmjason Yeah, I also thought about `Promise.all` but didn't try it :smile:

ycmjason:
@meteorlxy 
https://github.com/vuejs/vuepress/commit/9719bc3246afcf77f69eaf94a34f0d6b2fa9c95b

Looks like the for-await loop is intentional. I wonder why tho.

ycmjason:
I see, this is probably for better CLI output. The time is not much different so Evan chose to use for-await probably..

Feofilakt:
Why is it closed?

PR #316, opened 2018-05-01
fix(navbar): fix styles when navbar is not present

Aterbonus:
## Before
dektop guide before
mobile guide sidebar before
## After
dektop guide after
mobile guide sidebar after

### DIFF
--- a/lib/default-theme/styles/mobile.styl
+++ b/lib/default-theme/styles/mobile.styl
@@ -21,9 +21,13 @@ $mobileSidebarWidth = $sidebarWidth * 0.82
     transition transform .2s ease
   .page
     padding-left 0
-  .theme-container.sidebar-open
-    .sidebar
-      transform translateX(0)
+  .theme-container
+    &.sidebar-open
+      .sidebar
+        transform translateX(0)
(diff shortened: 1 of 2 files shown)


ulivz:
Can you provide the OS info?

Aterbonus:
When a page has no navbar (```navbar: false``` in the front matter) there is a gap where the navbar was.

This is especially evident when we use a Custom Layout and we don't want to have a navbar.

## Example:
### HomeLayout.vue
```html
<template>
<div class="container">
    <vue-nixie-clock></vue-nixie-clock>
</div>
</template>

<style scoped>
.container {
    align-items: center;
    background-color: #000;
    display: flex;
    height: 100vh;
    justify-content: center;
}

.nixie-clock {
    width: 800px;
}
</style>
```
### README.md
```
---
layout: HomeLayout
navbar: false
---
```
### Before
screenshot_20180502_023406

### After
screenshot_20180502_023738

meteorlxy:
Why not add some mark on the snapshots? I even can't find the differences at first...

image
image

ulivz:
Thanks for the work!

Issue #319, opened 2018-05-02
sidebar: auto has a bug


shaodahong:
I have a markdow file with multiple `##` and `###`. I added `sidebar: auto` and found that the sidebar can only display the first md file

shaodahong:
It is also related this #287

shaodahong:
I understand. but it is a bug

themeConfig.sidebar:
```
{
    title: 'test',
    collapsable: false,
    children: [
        '',
        'vue',
    ]
}
```

If i don't set `sidebar: auto`, the sidebar will show two articles （`index.html``vue.html`） that I configured in `themeConfig.sidebar`

If i set `sidebar: auto`, the sidebar only show one article, i don't know what this idea is

ulivz:
since you said you understand, so why set `sidebar: auto` here?

`sidebar: auto` should be used when the folder only has 1 md file.

PR #320, opened 2018-05-02
fix sidebar: auto bug

shaodahong:
#319


shaodahong:
> need more advice

Because the sidebarDepth is a maximum of `3`, there is no need to `resolveHeaders`, but later support nested sidebar need to be modified

I hope more support nested sidebar

ulivz:
Thanks for your contribution!

See: https://github.com/vuejs/vuepress/issues/319#issuecomment-385931378

BTW, You shouldn't compromise the functionality of `sidebar: auto` when you develop any new features.

PR #321, opened 2018-05-02
Change the required node version

whoan:
I was getting this error:

>$ node bin/vuepress dev docs  
SyntaxError: **Invalid regular expression: /(?<=(^|\/))(index|readme)\.md$/: Invalid group**  
    at Object.<anonymous> (/data/proyectos/blog/vuepress/lib/prepare.js:239:17)

The problem is that [RegExp lookbehind was implemented in V8 in version v6.2.103][look-behind-v8] and it [was just merged into the node engine in version 8.10][look-behind-node].

Check [this forum][forum] for more information.

[look-behind-v8]: https://chromium.googlesource.com/v8/v8.git/+/473a6f5b03651fde33fd7b3228a7dc4e38e24914
[look-behind-node]: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#8.10.0
[forum]: https://groups.google.com/forum/#!msg/v8-users/r-SN2yuKTL8/pfwrSuqSBQAJ

PR #323, opened 2018-05-02
Change regex to avoid lookbehind

whoan:
An alternative to PR #321.
Also related to PR #308 to fix issue #306.

This is to avoid updating node to 8.10 to support *Regex lookbehind*.

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -236,7 +236,7 @@ async function genComponentRegistrationFile ({ sourceDir }) {
   return `import Vue from 'vue'\n` + components.map(genImport).join('\n')
 }
 
-const indexRE = /^\/?([^/]+\/)*(index|readme)\.md$/i
+const indexRE = /(^|.*\/)(index|readme)\.md$/i
 const extRE = /\.(vue|md)$/
 
 function fileToPath (file) {
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -243,7 +243,7 @@ function fileToPath (file) {
   if (isIndexFile(file)) {
     // README.md -> /
     // foo/README.md -> /foo/
-    return '/' + file.replace(indexRE, '/$1')
+    return file.replace(indexRE, '/$1')
   } else {
     // foo.md -> /foo.html
     // foo/bar.md -> /foo/bar.html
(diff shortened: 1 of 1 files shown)


meteorlxy:
I think it's buggy.

Have you tested?

meteorlxy:
I have ever used this in #308:
```js
const indexRE = /(^|\/)(index|readme)\.md$/i

    // README.md -> /
    // foo/README.md -> /foo/
    const fileReplace = file.replace(indexRE, '')
    return '/' + ( fileReplace === '' ? '' : fileReplace + '/')
```

But I prefer to change a single line, so I chose lookbehind.

whoan:
I tested only with this input and it worked:

    '/foo/Gitter-Room-Index.md' -> '/foo/Gitter-Room-Index.md'
    '/Gitter-Room-Index.md' -> '/Gitter-Room-Index.md'
    'Gitter-Room-Index.md' -> 'Gitter-Room-Index.md'
    'README.md' -> '/'
    'foo/README.md' -> '/foo/'
    '/foo/README.md' -> '/foo/'

meteorlxy:
@whoan Test in `origin/master` but errors occurs.

You can consider my code above, which I had tested well.

whoan:
@meteorlxy I had forgotten to remove a (now) duplicated slash because I am appending it in the replacement.

meteorlxy:
@whoan 
I suggest you to test in official mater branch first.

Your:
- `'foo/bar/guide/Readme.md'` => `'/guide/'`

Expected:
- `'foo/bar/guide/Readme.md'` => `'/foo/bar/guide/'`

lib/prepare.js
```diff
@@ -236,14 +236,14 @@ async function genComponentRegistrationFile ({ sourceDir }) {
   return `import Vue from 'vue'\n` + components.map(genImport).join('\n')
 }
 
-const indexRE = /(?<=(^|\/))(index|readme)\.md$/i
+const indexRE = /^\/?([^/]+\/)*(index|readme)\.md$/i
```
meteorlxy: This may help:
```js
const indexRE = /(^\/?([^/]+\/)*)(index|readme)\.md$/i
```

meteorlxy: And maybe this is enough:
```js
const indexRE = /(^|.*\/)(index|readme)\.md$/i
```

ulivz: The second one looks good, I just tested for it, so let me leverage it.

Issue #331, opened 2018-05-02
Algolia Theme Customization


bayssmekanique:
The Algolia theme is not responding to the theme customizations.

Looking at the `Navbar.vue` file, it is importing the aliased `@AlgoliaSearchBox`.  Changing the import to `./AlgoliaSearchBox.vue` resolves the theming issues, but I'm not certain if there is a reason that it was using the alias that I'm not seeing?

I'm happy to make a PR for this if there is no objection to the change.

bayssmekanique:
As I look further, it appears that the intentions behind using the alias to be able to bypass the theme inclusion based on the `isAlgoliaSearch` attribute. 

https://github.com/vuejs/vuepress/blob/4bf56d75659d80a5545bca4a49b10c0761a73619/lib/webpack/createBaseConfig.js#L44-L46

Replacing the reference in the import will cause the style to always be bundled.  I'm not certain how to work around this and I foresee this happening in the future with similar modules that are optional dependencies so I don't know if having the user modify the import is necessarily the correct approach.

bayssmekanique:
Solution submitted in PR #334 .

ulivz:
Not clearly know what do you want to do.

bayssmekanique:
@ulivz, if you eject the theme and then customize the `AlgoliaSearchBox.vue` file, the changes are not picked up because the `@AligoliaSearchBox` alias was pointing to `default_theme` directory in the root of the application.  The PR #334 changes the alias to point to a relative path, so by default it will use the root `default_theme` location, but when ejected it will use the customized vue file instead.

ulivz:
Hmmm, so why not describe it by using a issue template at first 😅

if you want maintainers to help you to solve a problem or merge a pr, you should take us to your context first. I won't understand what you want to do util you said you use `eject`.

meteorlxy:
@bayssmekanique 
In my opinion, `eject` aims to help user build there custom theme. In your own theme, you can import AligoliaSearchBox by something like `@theme/AlgoliaSearchBox.vue` with the help of `@theme` alias.

bayssmekanique:
@meteorlxy, I'm not entirely following your point regarding the `@theme/AlgoliaSearchBox.vue`, as this is more addressing the `eject` condition, not necessarily theme overriding, which I think might be what you are addressing?

[...]

PR #334, opened 2018-05-02
fix: algolia custom theme (close: #331)

bayssmekanique:


ulivz:
This change doesn't make sense. `relative` path means the consumer component will rely on the relative position to the `AlgoliaSearchBox` component.

As doc said: 

> once you eject, you are on your own and won't be receiving future updates or bug fixes to the default theme even if you upgrade VuePress.

So for now, you should change the reference path at your eject files directly:

From:

```
import AlgoliaSearchBox from '@AlgoliaSearchBox'
```

To:

```
import AlgoliaSearchBox from '[Your path]'
```

Issue #318, opened 2018-05-01
Vue components created with JSX is not working in Markdown files


HandOfGod94:
[vuepress-error.zip](https://github.com/vuejs/vuepress/files/1965089/vuepress-error.zip)

- OS: Windows 10
- Node.js: v9.3.0
- VuePress: 0.8.4
- Browser: Google Chrome Version 66.0.3359.117 (Official Build) (64-bit)
- Local installation
- Package Manager used: yarn

I'm using VuePress for documenting my components. For demo, purpose I'm registering it globally in `enhanceApp.js` and using it `.md` files.

Attaching a reproducible project with a custom component.
I have used `yarn` to install dependecies. 
To reproduce:
1. Go to `src/components/MyComponent.js`.
2. Toggle the `return` statement to see the output.

To view docs: run `yarn docs:dev`

I guess the problem is that it `VuePress` doesn't have `babel-plugin-transform-vue-jsx`. So is this thing in roadmap, or will not be supported? Is there any other way to handle JSX in render function for VuePress

If I use `createElement` method directly with arguments, its rendering properly.
```javascript
export default {
  name: 'my-component',
    render (h) {
        return h('div', {}, this.$slots.default)
    }
}
```
Now in my `.md` files if I use
```html
<se-container>Hello World</se-container>
```
Its working correctly.
But if I use `JSX` in `render` method, then it's showing `ReferenceError: React is not defined` in console.
```javascript
export default {
    name: 'my-component',
    render (h) {
        return (
            <div>{this.$slots.default}</div>
        )
    }
}
```

![error](https://user-images.githubusercontent.com/9387023/39490979-5b2677fe-4da8-11e8-92b3-210174b2a220.JPG)

yyx990803:
VuePress currently does not support JSX. We will need to move to Babel for this.

HandOfGod94:
Hi @yyx990803,
Thanks for quick reply
Shouldn't this be documented? Somewhere stating the limitation or in ToDo list of VuePress docs.
I'll try to play around in my local setup with babel+VuePress, but is there any specific reason why it was not used?

ulivz:
Will release it at 0.9.0: #419 ❤️

twigs67:
Shouldn't this work with version 3? I'm running 3.1.1 and I'm still getting this error.

amravazzi:
Same here. I'm having problems importing components built with JSX.

PR #336, opened 2018-05-02
feat: Using Babel for Vue components inside md files (#318)

HandOfGod94:
Vue components now created with jsx is rendering inside `.md` files.
I have removed `buble-loader` as it will be redundant to use both.
Let me know if we need to check any specific things.


yyx990803:
Can you confirm async/await works with the setup?

HandOfGod94:
Hi @yyx990803,
Yes, Its working as per the expectation.
I tried with following component.
```javascript
export default {
  name: 'my-component',
  mounted () {
    this.getMsg()
  },
  data () {
    return {
      msg: 'Initial Content'
    }
  },
  methods: {
    getMsg: async function () {
      let promise = new Promise((resolve, reject) => {
        setTimeout(() => resolve("Hello World from async"), 5000)
      })
      let result = await promise;
      this.msg = result
    }
  },
  render (h) {
    return (
      <div>{this.msg}</div>
    )
  }
}
```
And then used it inside `.md` file for demo as
```html
<my-component />
```

HandOfGod94:
Size Comparison

|Application|Before|After|
|---|---|---|
|`VuePress` Docs(Build)| 1.14MB| 1.17MB|
|`CustomApp` Docs(Build)|146KB|168KB|
|`CustomApp with JSX` Docs(Build)|`failed`|168KB|

ulivz:
@HandOfGod94 Thanks for the great work.

PR #343, opened 2018-05-05
feat: Allow page overrides for the sidebar

mdaffin:
This makes it possible to override the sidebar on a per-page level by adding a frontmatter that takes a list of pages in the same format as the site level sidebar. For example:

```
---
sidebar: 
  - /config/
  - /default-theme-config/
---
```

### DIFF
--- a/lib/default-theme/util.js
+++ b/lib/default-theme/util.js
@@ -119,7 +119,7 @@ export function resolveSidebarItems (page, route, site, localePath) {
     ? themeConfig.locales[localePath] || themeConfig
     : themeConfig
 
-  const sidebarConfig = localeConfig.sidebar || themeConfig.sidebar
+  const sidebarConfig = pageSidebarConfig || localeConfig.sidebar || themeConfig.sidebar
   if (!sidebarConfig) {
     return []
   } else {


ulivz:
I think each page's sidebar can only set links with its own fs-level's children files. or, you might put the file in the wrong position.

ulivz:
This usage is inconsistent to the default usage, user will be very confused why two usages are different.

> This looks to be an existing bug, not one introduced by this change.

Are you sure you have understand the code?

ulivz:
You can open an issue with this feature request. we will plan for that.

mdaffin:
The usage is consistent - I see the same behaviour if you place the settings in config.js or in the pages frontmatter. When I added `sidebar: ['/guide/assets/']` to `config.js` it breaks for the same reason as when you place it in the front matter. No behaviour change here.

Both using either `/guide/assets` or `/guide/assets.html` works in both cases as well. There is no behaviour difference and nothing to be confused about.

`/guide/assets/` seems to be an existing bug that we can address separately as it does not seem to be a regression with this change.

How exactly does this usage change from the existing behaviour? Even the nested grouping works as expected [like here](https://raw.githubusercontent.com/mdaffin/disconnected.systems/master/site/blog/archlinux-installer/index.md).

PR #345, opened 2018-05-05
feat: register globally theme components (close #281)

### LINKED ISSUE #281: Register global components in themes
Vue Press version: 0.8.4

In `.vuepress/` components are automatically registered when they're placed in `components/`. As a developer, my assumption was that the `theme/` folder would have a similar feature.

Components can be registered with a pattern like this:
```javascript
//.vuepress/theme/enhanceApp.js

import PostList from './components/PostList'
import PostSnippet from './components/PostSnippet'

export default ({
    Vue, // the version of Vue being used in the VuePress app
    options, // the options for the root Vue instance
    router, // the router instance for the app
    siteData // site metadata
  }) => {
    Vue.component('PostList', PostList)
    Vue.component('PostSnippet', PostSnippet)
  }
```

But it would be nice if the theme could mirror functionality of `.vuepress/` structures like `components/`

hmatalonga:
This pull request allows to register theme components globally, just like the components in `.vuepress/components`.

It makes use of `options.themePath` to include all *.vue in a `components` sub-folder to follow the existing logic.

Closes #281

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -245,16 +245,22 @@ async function resolveOptions (sourceDir) {
   return options
 }
 
-async function genComponentRegistrationFile ({ sourceDir }) {
-  function genImport (file) {
+async function genComponentRegistrationFile ({ sourceDir, themePath }) {
+  function genImport (file, ...componentDir) {
     const name = fileToComponentName(file)
-    const baseDir = path.resolve(sourceDir, '.vuepress/components')
+    const baseDir = path.resolve(...componentDir)
     const absolutePath = path.resolve(baseDir, file)
     const code = `Vue.component(${JSON.stringify(name)}, () => import(${JSON.stringify(absolutePath)}))`
     return code
   }
-  const components = (await resolveComponents(sourceDir)) || []
-  return `import Vue from 'vue'\n` + components.map(genImport).join('\n')
+  const components = (await resolveComponents(sourceDir, '.vuepress/components')) || []
+  const themeComponents = (await resolveComponents(themePath, 'components')) || []
+  const componentsStr =
+      components.map(f => genImport(f, sourceDir, '.vuepress/components')).join('\n')
+  const themeComponentsStr =
+      themeComponents.map(f => genImport(f, themePath, 'components')).join('\n')
+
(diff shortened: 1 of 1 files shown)


meteorlxy:
For me, I prefer to put the components of my custom theme in the `themePath/components` directory, rather than register them as global components.

What about `themePath/globalComponents` ?

I don't know which solution is better, may need more discussions.

hmatalonga:
Sure thing, we should hear about the general consensus which name to give the folder :+1:

yliaho:
I agree with @meteorlxy – `globalComponents` (or similar) would be preferable.

hmatalonga:
Giving it some more thought, I agree it makes perfectly sense to have a dedicated `globalComponents` where only those ones are registered.

I can make that change, if you feel like it's an adequate solution for everyone :+1:

SebastianNiemann:
Regarding the path name, i would vote for either for `components ` or to rename `.vuepress/components` to `.vuepress/globalComponents`.

Differentiating the folder name between `.vuepress` and a theme, while serving the same purpose, seems to be an avoidable learning curve increase.

I like to think about themes as a common `.vuepress` functionality for multiple sites. Therefore being as closely related to each other as possible.

AuditorCJ:
Giving it some more thought, I agree it makes perfectly sense to have a dedicated globalComponents where only those ones are registered.

I can make that change, if you feel like it's an adequate solution for everyone +1

hmatalonga:
@AuditorCJ 

It was decided to wait until the plugin API is released, to add this feature as a plugin then we will revisit this discussion and decide on which directory naming to follow :wink:

ulivz:
A theme's structure would be:

```
theme
├── global-components
│   └── xxx.vue
├── components
│   └── xxx.vue
├── layouts
│   ├── Layout.vue (必要的)
│   └── 404.vue
├── styles
│   ├── index.styl
│   └── palette.styl
├── index.js
├── enhanceApp.js
└── package.json
```

And those components under `global-components` will be registered as global components.

ref: https://v1.vuepress.vuejs.org/theme/writing-a-theme.html#directory-structure

PR #354, opened 2018-05-07
feat: support multiple dev servers by using random temp directories

meteorlxy:
Current **global** vuepress does not support multiple instances.

In other words, we cannot open multiple `dev` server using global installed vuepress.

The reason is we are now using a **fixed**/**single** temp directory, i.e. `lib/app/.temp`. Each project will overwrite the same temp directory, which will cause collision.

For enhancement, I introduce [tmp](https://www.npmjs.com/package/tmp) module, which is widely used by many projects.


yarn.lock
```diff
@@ -6222,7 +6222,7 @@ tiny-emitter@^2.0.0:
 
 tmp@^0.0.33:
   version "0.0.33"
-  resolved "https://registry.yarnpkg.com/tmp/-/tmp-0.0.33.tgz#6d34335889768d21b2bcda0aa277ced3b1bfadf9"
+  resolved "http://registry.npm.taobao.org/tmp/download/tmp-0.0.33.tgz#6d34335889768d21b2bcda0aa277ced3b1bfadf9"
```
ulivz: This change shouldn't be included here.

meteorlxy: Oh, forget to add `-g` when change the registry

meteorlxy:
@ulivz 

Changed to local package dir.

ycmjason:
I wonder why would we need to spawn multiple dev servers?

ulivz:
Same question, why would we need to spawn multiple dev servers?

meteorlxy:
Good question, why hexo & jekyll support multiple dev servers?

ulivz:
Hmmm, do you mean that we must support others' support? Where is the actual meaning?

meteorlxy:
LOL, don't mind my word.

Users may have different requirements. For me, I'm `dev`ing the official repo and my own custom theme in the mean time.

Or I have another question: why we restrict users to open only one dev server?

hmatalonga:
@meteorlxy 

While it is not strictly necessary... I agree it can be handy to have that possibility :smiley:

ulivz:
Shouldn't it be a feature request first? If you just submitted a PR and hoped to be merged as soon as possible, then we will discuss whether we need to do it. Will you put the cart before the horse?

BTW, I don't want and DON'T have time to argue with you at all. If you want to support a feature, please submit a feature request first.

meteorlxy:
I did, sir #355... And you tagged it as `enhancement` 😓

ulivz:
And, can you confirm that will the old temp file be deleted when reboot?

meteorlxy:
@ulivz They will be deleted immediately after the process exits.

However, there may be some exceptions that make the process interrupt unexpectedly, and the temp files will remain there.

If we use system tmp directory as my first commit, we don't have to think about it.

If we use local package directory as you suggested, it could be a potential problem.

ulivz:
We can simply implement it with native node.js. since we didn't need `tmp`'s other features.

And delete the `.temp-*` dirs before `prepare`.

meteorlxy:
After discussion on slack, so for we (@ulivz , @ycmjason and I) have some conclusion:
- Multiple dev servers is necessary in these possible scenarios:
  - Developing multiple vuepress projects in the mean time, or trying to campare some changes with each other.
  - Multiple users access to one remote machine, and start vuepress dev server in the mean time.
- Although `tmp` module can cleanup the tempfile after the process exits, if the vuepress process was killed by accident, the temp files will still remain there. So we would prefer to store the temp files to OS's temp dir. (But we cannot determine if this is a good solution. cc @yyx990803 )

Issue #186, opened 2018-04-20
Feature request: make `target=_blank rel=…` for external links optional


mathiasbynens:
https://github.com/vuejs/vuepress/blob/4cf123240322898c44cab0ada5a729ca66ba9094/lib/markdown/link.js#L16

`target=_blank` denies people the opportunity to open the link in the same window. Please consider making it possible to disable the automatic addition of these attributes.

lbenie:
would it be best to add an option to the base config ?

ulivz:
@mathiasbynens 

We can support following format to add custom attributes for external link:

```md
[link-text]((querystring)link)
```

For example:

```md
[custom domain]((target=_self&class=my-link)http://surge.sh/)
```

It will be rendered as:

```
<a href="http://surge.sh/" target="_self" class="my-link">surge</a>
```

cc @yyx990803

ycmjason:
I have created two PR with two approaches. Perhaps we need to decide which way we like more. I personally like #357 more as it is less low-level.

ycmjason:
@ulivz please close this as #358 is merged.

ulivz:
Fixed at 20e5bd89540d13af9426967eae84e30a97c7b65b, thank for @ycmjason

PR #356, opened 2018-05-07
feat: support custom attributes for external links in markdown

ulivz:
# Summary

- Close: #186

This PR is to support following format to add custom attributes for ONLY **external links**, internal link will keep the old behaviors.

```md
[link-text]((querystring)link)
```

## example 1

- Input: 

```md
[custom domain]((target=_self&class=my-link)http://surge.sh/)
```

- Output: 

```
<a href="http://surge.sh/" target="_self" rel="noopener noreferrer" class="my-link">custom domain</a>
```

## example 2

```md
[custom domain]((target=&rel=)http://surge.sh/)
```

It will be rendered as:

```
<a href="http://surge.sh/">custom domain</a>
```

## example 3

```md
[custom domain](http://surge.sh/)
```

It will be rendered as:

```
<a href="http://surge.sh/" target="_blank" rel="noopener noreferrer">surge</a>
```

So this PR is fully downward compatible.

cc @meteorlxy @ycmjason

### DIFF
--- a/lib/markdown/link.js
+++ b/lib/markdown/link.js
@@ -1,6 +1,7 @@
 // markdown-it plugin for:
 // 1. adding target="_blank" to external links
 // 2. converting internal links to <router-link>
+const querystring = require('querystring')
 
 module.exports = md => {
   let hasOpenRouterLink = false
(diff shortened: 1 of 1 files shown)


ulivz:
Will add doc later.

ycmjason:
Just my opinion 😄 

I understand that you want to make markdown links more flexible 👍, but I have a feeling that this makes things a little bit too complex? If I were to type the `querystring`, it is probably easier for me to type html:

```html
<a href="https://github.com">hi</a>
```

I think this feature request is legitimate and in my opinion the best solution would be having a global config that switches all links. For special cases, just let the user type `<a>` tags. 

What do you think?

mathiasbynens:
I think my request in #186 is entirely orthogonal to this proposal.

To clarify, #186 is asking for a way to disable the automatic addition of `target="_blank" rel="…"` to external links. I don’t need fine-grained control over attributes on links. I just want some kind of configuration setting that I can toggle to stop these attributes from appearing, without having to change my Markdown source files.

ycmjason:
I have created two PR with two approaches. Perhaps we need to decide which way we like more. I personally like #357 more as it is less low-level.

ulivz:
@ycmjason It's truly making things more complicated, I agree with your opinion 👍

@mathiasbynens Excuse me, I just misunderstood your proposal and thought that you might need fine-grained control 😅

Closing it and go #357 to continue.

PR #357, opened 2018-05-07
Added global config for external links

ycmjason:
Fixing https://github.com/vuejs/vuepress/issues/186

This approach adds a config which switches all external links to use `blank` or no `blank`.

Alt approaches: https://github.com/vuejs/vuepress/pull/356 https://github.com/vuejs/vuepress/pull/358

cc @ulivz @meteorlxy @mathiasbynens

### DIFF
--- a/docs/guide/markdown.md
+++ b/docs/guide/markdown.md
@@ -48,10 +48,12 @@ Given the following directory structure:
 
 ### External Links
 
-Outbound links automatically gets `target="_blank"`:
+Outbound links automatically gets `target="_blank" rel="noopener noreferrer"`:
 
 - [vuejs.org](https://vuejs.org)
 - [VuePress on GitHub](https://github.com/vuejs/vuepress)
+
+You could alter this this behavior by setting `config.markdown.link.blank = false`. See more [here](/config/#markdown-link).
 
 ## Front Matter
 
--- a/lib/markdown/link.js
+++ b/lib/markdown/link.js
@@ -1,8 +1,9 @@
 // markdown-it plugin for:
-// 1. adding target="_blank" to external links
+// 1. adding target="_blank" and rel="noopener noreferrer" to external links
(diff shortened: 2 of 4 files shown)


lib/markdown/link.js
```diff
@@ -14,8 +15,10 @@ module.exports = md => {
       const isExternal = /^https?:/.test(href)
       const isSourceLink = /(\/|\.md|\.html)(#.*)?$/.test(href)
       if (isExternal) {
-        addAttr(token, 'target', '_blank')
-        addAttr(token, 'rel', 'noopener noreferrer')
+        if (opts.blank) {
```
ulivz: `blank` should only control the `target`, and `noopener` and `noreferrer` are both worth to be added to external links by default.

See: [Opens External Anchors Using rel="noopener"](https://developers.google.com/web/tools/lighthouse/audits/noopener) 

> Adding a rel="noopener" attribute prevents the new page from being able to access the window.opener property and will ensure it runs in a separate process. The rel="noreferrer" attribute has the same effect, but will also prevent the Referer header from being sent to the new page. See HTML Standard: Link type "noreferrer" for an explanation of this behavior.

Also see: [The performance benefits of rel=noopener](https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/)

> If you have links to another origin, you should use rel="noopener".

docs/guide/markdown.md
```diff
@@ -48,11 +48,13 @@ Given the following directory structure:
 
 ### External Links
 
-Outbound links automatically gets `target="_blank"`:
+Outbound links automatically gets `target="_blank" rel="noopener noreferrer"`:
```
ulivz: `rel="noopener noreferrer"` should be remained here.

ulivz:
Thanks for the great work! 👍Closing it and leverage #358

PR #358, opened 2018-05-07
Global config for links attr

ycmjason:
Fixing https://github.com/vuejs/vuepress/issues/186

Alternative approach to https://github.com/vuejs/vuepress/pull/356 and https://github.com/vuejs/vuepress/pull/357.

This approach added `config.markdown.externalLinks` which is an object whose keys and values will be added as a attributes to external links.

cc @ulivz @meteorlxy @mathiasbynens


ulivz:
Will release it at 0.9.0. #419. 🚀

PR #361, opened 2018-05-08
Avoid duplicate slash when docsDir is not set

whoan:
When `docsDir` is not set, the url to edit the document currently has two slashes.


ulivz:
Will release it at 0.9.0: #419. 🚀

PR #375, opened 2018-05-09
学习使用，并对config.js进行标注

bignumlock:


ulivz:
Thanks, but it shouldn't be included in this project.

Issue #370, opened 2018-05-08
NotFound.vue missing title and description


bayssmekanique:
- VuePress `0.8.4`
- Tested on Windows 10 & Ubuntu with same result

There seems to be 2 different ways this issue appears, depending on if you are running dev or a static build.

When in development mode (`vuepress dev`) the title attribute is blank and the description attribute is missing entirely.

When building the site (`vuepress build`) the resulting `404.html` page has `VuePress` as the title and the description attribute is blank.

Also worth noteing, head tags listed in the `head` attribute of property of the `config.js` file are appearing in the head correctly.

bayssmekanique:
This appears to be the culprit: 
https://github.com/vuejs/vuepress/blob/272df57421d20907b5a1edecabcaad649641e791/lib/build.js#L72-L75

https://github.com/vuejs/vuepress/blob/272df57421d20907b5a1edecabcaad649641e791/lib/build.js#L133-L171

hmatalonga:
I also believe that this is related too.

https://github.com/vuejs/vuepress/blob/272df57421d20907b5a1edecabcaad649641e791/lib/build.js#L18

Since the sourceDir is passed to `prepare.js` the existing pages will make use of `inferTitle(frontmatter)` to generate the title, I am guessing. So because the 404 default page is added afterwards, that is not applied on that page, therefore the defaults title and description will be used without inferring the proper title...

https://github.com/vuejs/vuepress/blob/272df57421d20907b5a1edecabcaad649641e791/lib/prepare.js#L202-L217

Maybe this will help detecting the issue :stuck_out_tongue:

meteorlxy:
See #349 , same cause

hmatalonga:
@meteorlxy 

I just saw the other issues related to this. I hadn't seen your PR #314 yet, will it be merged? that could probably solve all theses cases. Or you are waiting until the Plugin API is published?

meteorlxy:
@hmatalonga I think #314 could be merged at current stage, as it can solve some problems somehow when the plugin API has not published.

What do you think about that @ulivz

hmatalonga:
I was wondering also if there will be a minor release before the plugin API? Since latest merged PRs introduced new things to the docs...

meteorlxy:
@hmatalonga 

Evan is busy these days and doesn't have time to look at this project. We have no auth to release a new version of npm package now.

meteorlxy:
@hmatalonga I create another PR #376 for this.

[...]

Issue #349, opened 2018-05-06
Title and description in custom theme


pavloko:
- Your OS: MacOS 10.13.4
- Node.js version : 9.4.0
- VuePress version : 0.8.4
- Is this a global or local install? : global install
- Which package manager did you use for the install? : npm

When using a custom theme via `.vuepress/theme/Layout.vue`, config variables `title` and `description` specified in `.vuepress/config.js` are not inserted in the `dist/index.html` file.

ulivz:
Can you provide the repro repo?

meteorlxy:
It's implement in default theme, not in core.

Copy the related code from `Layout.vue` of default theme.

(So I prepose to extract them into mixins for better reuse #314 . Consider merge it? @ulivz :smiley: )

whoan:
@pavloko Do you have any locale set?

Notice that if you have something like this:

    title: 'Hello',
    description: 'World',
    locales: {
      '/': {
        lang: 'en-US',
        title: 'VuePress',
        description: 'Vue-powered Static Site Generator'
      },
      '/zh/': {
        lang: 'zh-CN',
        title: 'VuePress',
        description: 'Vue 驱动的静态网站生成器'
      }
    }

**The locale will have precedence and that may be the reason of your problem**...

It is worth mentioning that with the field `description` it even has more precedence any value in a page's *Front Matter*.

ulivz:
No any reproduced link, all the "conclusions" will be inconclusive.

meteorlxy:
https://github.com/vuejs/vuepress/blob/5e87b657499fbe44d6bd71bdb6b343e41faa6b5e/lib/default-theme/Layout.vue#L90-L96

meteorlxy:
@pavloko Yes, that's the point, so I mentioned:

> Maybe we should put the code above to core, as it's necessary for the ssr template:

I would create a PR for that, but there is another problem, see #296 .

In #296, I wonder whether should we restrict `$title` in core.
In this issue, we are going to also restrict `this.$ssrContext.title = this.$title` in core. Uh what a mess!

pavloko:
Hey @meteorlxy! Sorry, missed that line between the code blocks. I shared my thought in #296. 

I'd agree that any `head` modifications should belong to core/config.

ulivz:
@pavloko Now the `title`, `description` and `lang` will be rendered at not only default theme but also custom themes.

[...]

PR #376, opened 2018-05-09
fix: move ssrContext setting to core (#349, #370)

meteorlxy:
- `$ssrContext` is necessary for SSR, so it should be in the core, not in the default theme. #349 
- `NotFound.vue` doesn't match any `$page`, so I provide an empty page data for it in `lib/app/util.js/findPageForPath`, or errors will occur when rendering it. #370 
- As `dataMixin` is not only `data` after this PR, consider renaming it to `globalMixin`? @ulivz

### DIFF
--- a/lib/app/dataMixin.js
+++ b/lib/app/dataMixin.js
@@ -80,5 +80,12 @@ export default {
         this.$route.path
       )
     }
+  },
+  created () {
+    if (this.$ssrContext) {
+      this.$ssrContext.title = this.$title
+      this.$ssrContext.lang = this.$lang
+      this.$ssrContext.description = this.$page.description || this.$description
+    }
   }
 }
--- a/lib/app/util.js
+++ b/lib/app/util.js
@@ -13,4 +13,8 @@ export function findPageForPath (pages, path) {
       return page
     }
   }
+  return {
+    path: '',
+    frontmatter: {}
+  }
 }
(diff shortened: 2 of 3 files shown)


lib/app/dataMixin.js
```diff
@@ -80,5 +80,12 @@ export default {
         this.$route.path
       )
     }
+  },
+  created () {
```
ulivz: It shouldn't be placed here, or it will be invoked for each component.

lib/app/util.js
```diff
@@ -13,4 +13,8 @@ export function findPageForPath (pages, path) {
       return page
     }
   }
+  return {
+    path: '',
+    frontmatter: {}
+  }
```
ulivz: Is it to fix some bugs, or?

lib/app/dataMixin.js
meteorlxy: If we don't put it here, we have some other choices:
1. Ask users to put it mannually into the custom `Layout.vue`
2. Wrap the `themeLayout` with another components, to which inject this snippets.

lib/app/util.js
meteorlxy: Without this, there will be errors rendering `NotFound.vue`, as there is not `$page` match it.

meteorlxy: In fact it to fix a bug. I'll extract it to another PR

meteorlxy: #388

Issue #365, opened 2018-05-08
[feature request] code block shows line number


xkcoding:
I hope add a feature.

make the code block to show line number.

david-trejo-ck:
Ideally if you add this, the line number won't be copy-pasteable when you copy-paste the code :)

xkcoding:
yep

PR #379, opened 2018-05-09
feat: code line number (close: #365)

ulivz:
cc @egoist , it also fixed a empty line highlight issue:

```js
`<span class="highlighted-line">${split || '\n'}</span>`
```

# Snapshot

## Highlight line + Line number

image

## Only line number

image

## Only Highlight line

image


ycmjason:
I think it would be nicer to make this a separate markdown-it plugin. Semantically it is not right to put this feature in highlightLines plugin.

P.S. I am working on the same thing too but haven't yet complete it.

ycmjason:
Also, the current implementation make the line number selectable? Which might not be ideal?

ulivz:
> 1. make this a separate markdown-it plugin

Sure, I originally make it as a separate plugin, but since `markdown-it-highlight-lines` rewrite the render method, so it won't work.

> 2. the current implementation make the line number selectable

have handled for that.

xkcoding:
@ulivz Unbelievable. That's what I want.

ulivz:
# Updated 5-13

## Desktop

line-numbers

## Mobile
line-numbers-mobile

PR #382, opened 2018-05-09
fix(style): sets max-width 100% for images in .content

### LINKED ISSUE #381: Content Layout Bug: Image Runs Out Horizontally
Found a layout issue. When an image is wide enough, it will "run out" of the content area and make the page horizontally scroll-able, primarily visible on mobile.

refs: edm00se/awesome-board-games#5

edm00se:
An overly wide hero image will run out of the container and make the otherwise normally sized page
on mobile horizontally scroll-able.

refs: edm00se/awesome-board-games#5

fixes #381

### DIFF
--- a/lib/default-theme/Home.vue
+++ b/lib/default-theme/Home.vue
@@ -97,6 +97,9 @@ export default {
       color lighten($textColor, 10%)
     p
       color lighten($textColor, 25%)
+  .content
+    img
+      max-width 100%
   .footer
     padding 2.5rem
     border-top 1px solid $borderColor


edm00se:
😁

ycmjason:
Well actually, we have this already at

https://github.com/vuejs/vuepress/blob/7813fa8fbfefa6c8d2165a1c0fc584b7c5d46717/lib/default-theme/styles/theme.styl#L68

The problem is probably because we have used `<Content custom/>` in `Home.vue`.

https://github.com/vuejs/vuepress/blob/7813fa8fbfefa6c8d2165a1c0fc584b7c5d46717/lib/default-theme/Home.vue#L19

Not sure about the reason behind this decision. We probably need to wait for @ulivz for advice. :smile:

edm00se:
Curious, as the result wasn't observable, was it included in the `0.8.4` version?

ycmjason:
@edm00se 
As I mentioned, this is because we have `<Content custom/>` which appends `custom` class to the container. So the style is not pulled through. Currently this style only apply to page contents.

edm00se:
Aha. Well, this ought to take care of it on the other side of things then.

edm00se:
Ah, I see, since it's already split logic. I'll close this PR and work up a new commit for the `.content.custom` section.

Thanks.

PR #388, opened 2018-05-10
fix: computed parsing errors in notfound

meteorlxy:
# Before
before

# After
after

# Cause
No `page` matches `NotFound.vue`, so we have to provide an empty page object


ulivz:
Will release it at 0.9.0: #419. 🚀

PR #389, opened 2018-05-10
fix: move ssrContext and updateMeta to core (#349, #370) 

### LINKED ISSUE #349: Title and description in custom theme

### LINKED ISSUE #370: NotFound.vue missing title and description

meteorlxy:
Another implementation of #376 

fix #349 and fix #370

### DIFF
--- a/docs/zh/default-theme-config/README.md
+++ b/docs/zh/default-theme-config/README.md
@@ -327,7 +327,7 @@ module.exports = {
     docsDir: 'docs',
     // 假如文档放在一个特定的分支下：
     docsBranch: 'master',
-    // 默认是 true, 设置为 false 来禁用
+    // 默认是 false, 设置为 true 来启用
     editLinks: true,
     // 默认为 "Edit this page"
     editLinkText: '帮助我们改善此页面！'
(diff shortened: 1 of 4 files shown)


ulivz:
Thanks, but for such function, we don't need to create a new component to increase the burden of the client, and should do some compilation time optimization instead.

Fixed at fcaee80358d47a26753eae49979a2047757a1fd8

Issue #369, opened 2018-05-08
Use of "docs" in Custom Themes page


cfjedimaster:
On the Custom Themes page (https://vuepress.vuejs.org/guide/custom-themes.html), this text is mentioned:

"The `Layout` component will be invoked once for every .md file in `docs`"

Except what is `docs`? Did you mean simply the root of the VuePress project?

hmatalonga:
All the pages of your project will make use that Layout component.

Say you have this project structure:
```
+-- README.md
+-- guide/
|   +-- intro.md
|   +-- something-else.md
```

All three files *.md would be rendered based on your custom theme Layout component.

`docs` is simply referring to the case of Vuepress documentation itself which has its root folder named docs, like you said :wink:

ulivz:
> what is `docs`?

The `docs` means your `targetDir`, execute `vuepress -h` to see it:

```
  Usage: vuepress <command> [options]

  Options:

    -V, --version                output the version number
    -h, --help                   output usage information

  Commands:

    dev [options] [targetDir]    start development server
    build [options] [targetDir]  build dir as static site
    eject [targetDir]            copy the default theme into .vuepress/theme for customization.
```

cfjedimaster:
I see that you closed the PR and that you commented that the original already makes sense. Respectfully I disagree. The docs here were confusing and saying the CLI clears it up doesn't address the issue in the actual docs themselves. May I ask we re-open this, or reconsider the PR as it did seem to clear it up a bit.

PR #390, opened 2018-05-10
docs: make the docs root specific

meteorlxy:
#369


ulivz:
I think the original is already making sense...

BTW, the `sourceDir` is a new word to end user since it's only existed at our source code.

Issue #392, opened 2018-05-10
Global vuepress fail to resolve custom theme


meteorlxy:
Here's the simplest way to repro it:

1. Add a custom theme (take `vuepress-theme-vue` for example) to a new repo:
```
mkdir testrepo
cd testrepo
yarn add vuepress-theme-vue
```

2. Create `src` and base files, then the directory will be like this:

```
+-- src/
|   +-- .vuepress/
|       +-- config.js
|   +-- README.md
+-- node_modules
|   +-- vuepress-theme-vue/
+-- package.json
+-- yarn.lock
```

3. Set theme to `vue` in `config.js`

```js
module.exports = {
  theme: 'vue'
}
```

4. Run global `vuepress`
```
vuepress dev src
```

5. Error
```
Error: [vuepress] Failed to load custom theme "vue". File vuepress-theme-vue/Layout.vue does not exist.
```

6. Possible solutions

- Install custom theme globally (i.e. `yarn global add vuepress-theme-vue`）
- (Or) Install `vuepress` locally (i.e. `yarn add vuepress`) and run it locally (i.e. `yarn vuepress dev src`)

The cause is that we are using `require.resolve()` to resolve custom theme:

https://github.com/vuejs/vuepress/blob/145cf4f244a9b2fff3ee9a3c59985dcda9642aa6/lib/prepare.js#L139-L140

Is there a better solution?

meteorlxy:
@hmatalonga Come on!

Look at the **last line of the first step** to repro it:
> yarn add vuepress-theme-vue

hmatalonga:
@meteorlxy my apologies, I don't known how I how missed the first steps... I removed my dumb comments from the previous message xD

Anyway for using it via global vuepress does the idea of somehow fetching the theme contents makes sense to you?

meteorlxy:
@hmatalonga 
As a global `cli`, it should work, I think.

In my opinion, we must consider every possible & sensible scenario

ulivz:
> In my opinion, we must consider every possible & sensible scenario

once again, I want to emphasize the intention of Evan: **VuePress is mainly for vuejs**. 

So we don't have to think about all possible scenarios.

ulivz:
I don't think this feature is sensible. If you just want to play around with VuePress, you can install it globally. but since you have a custom theme installed at local, why not installing it at local too?

meteorlxy:
"If you just want to play around with VuePress" - and try some different themes randomly - this is sensible. 

[...]

PR #399, opened 2018-05-10
fix: resolve custom theme from global cli (#392)

meteorlxy:
Details in #392

Reference [Modules | Node.js Documentation](https://nodejs.org/api/modules.html#modules_require_resolve_request_options)


meteorlxy:
Emmm...What if:
```
+-- node_modules/
+-- Readme.md
+-- hello-ulivz.md
+-- package.json
+-- yarn.lock
....
```

It's not good but something like our [As Easy as 1, 2, 3](https://vuepress.vuejs.org/#as-easy-as-1-2-3)

meteorlxy:
Hmmm you're right. `path.resolve(__dirname, '../node_modules')` could be better?

PR #400, opened 2018-05-10
fix: glob patterns error on windows (#348)

### LINKED ISSUE #348: On Windows, the chokidar watcher does not work well
- OS: Windows 10 x64
- Node v8.11.1

Related to #346 , not only components, but also markdown files.

The current setting of `chokidar` only tracks the existed files, but does not track new files on **Windows**. (Maybe this is an issue of `chokidar`)

https://github.com/vuejs/vuepress/blob/d9b290b39863c51f60468b7091d82f52a6c7335a/lib/dev.js#L30-L41

In addtion, unlinking a markdown/component file will cause an webapck error first, and then successfully compiled after the `prepare` finish. Could be enhanced somehow?

meteorlxy:
fix #348 

It's not an issue of `chokidar` but ours, as the glob patterns only work with `/` not `\`.
On Windows, `path` module will resolve `/` to `\`.

https://github.com/paulmillr/chokidar/issues/714

I misunderstood what he meant and made some meaningless conversation... I just want to express my thanks to the guy but he locked it immediately 😅

### DIFF
--- a/lib/dev.js
+++ b/lib/dev.js
@@ -26,8 +26,7 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
       console.error(chalk.red(err.stack))
     })
   }
-  console.log(path.join(sourceDir, '**/*.md'))
-  console.log(path.resolve(sourceDir, '**/*.md'))
+
   // watch add/remove of files
   const pagesWatcher = chokidar.watch([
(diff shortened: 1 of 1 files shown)


meteorlxy:
@ulivz 🤣 Oh you didn't see it

PR #410, opened 2018-05-11
feat: allow themes to set default themeConfig

### LINKED ISSUE #409: [feature request] Allow themes to set default themeConfig
A theme may need some basic config to make it run, or some default config that don't need to be set mannully by users.

So I propose to allow themes to set default `themeConfig`, which can be overridden in `siteConfig.themeConfig`.

I've put it together with another feature in PR #315 , which has not been approved yet. But I think theme default config could be an essential feature for a theme maker, so I decide to extract it to a single PR #410.

meteorlxy:
close #409


ulivz:
https://github.com/vuejs/vuepress/issues/409#issuecomment-388371890

PR #425, opened 2018-05-12
fix: upgrade webpack-serve and avoid port conflict

### LINKED ISSUE #424: The error message of port in use is not correct. Please check whether it is hard-coded.
In my laptop, port 8080 is in use, port 8081 is free. 

When I start `vuepress dev` command without any further configuration, it said port 8081 is in use. **I think it should be 8080, and perhaps this error message is hard-coded. Please check it.**

```
D:\JavaScript\DemoVuePress>vuepress dev
Extracting site metadata...
Compiling

* Client █████████████████████████ compiling (0%)


× ｢serve｣:  { Error: listen EADDRINUSE 127.0.0.1:8081
    at Object._errnoException (util.js:1024:11)
    at _exceptionWithHostPort (util.js:1046:20)
    at Server.setupListenHandle [as _listen2] (net.js:1351:14)
    at listenInCluster (net.js:1392:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1501:7)

  VuePress dev server listening at http://localhost:8081/
```
According to the vuepress documents, the default port is 8080, which is not free in my laptop, so I changed the default port to anther free port, everything is okay.

## Working Environment ##
- Windows 7
- Node.JS 8.9.3
- VuePress 0.8.4

meteorlxy:
close #424

related https://github.com/webpack-contrib/webpack-serve/issues/135

## Before
---
before

## After
---
after

### DIFF
--- a/lib/dev.js
+++ b/lib/dev.js
@@ -108,7 +108,7 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
     host,
     dev: { logLevel: 'warn' },
     hot: {
-      port: parseInt(port) + 1,
+      port: port + 1,
       logLevel: 'error'
     },
     logLevel: 'error',
--- a/lib/dev.js
+++ b/lib/dev.js
@@ -107,7 +107,10 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
     compiler,
     host,
     dev: { logLevel: 'warn' },
-    hot: { logLevel: 'error' },
+    hot: {
+      port: parseInt(port) + 1,
+      logLevel: 'error'
+    },
     logLevel: 'error',
     port,
     add: app => {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -95,7 +95,7 @@
     "webpack": "^4.8.1",
     "webpack-chain": "^4.6.0",
     "webpack-merge": "^4.1.2",
-    "webpack-serve": "^0.3.1",
+    "webpack-serve": "0.3.2",
```
ulivz: Don't forget `^`

lib/dev.js
```diff
@@ -107,7 +107,10 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
     compiler,
     host,
     dev: { logLevel: 'warn' },
-    hot: { logLevel: 'error' },
+    hot: {
+      port: parseInt(port) + 1,
```
ulivz: `portfinder.getPortPromise` will return `number`, so `parseInt ` is unnecessary here.

PR #426, opened 2018-05-12
added outboundlink icon to nav external link

ycmjason:
Currently the `OutboundLink.vue` is used only for Github nav. Would be nice if other external link also get it.


ycmjason:
Will open another PR fixing more stuff

PR #447, opened 2018-05-14
vue-analytics + enhanceApp.js example

brandonkelly:
The [vue-analytics](https://github.com/MatteoGabriele/vue-analytics) plugin just added support for VuePress. Considering most VuePress sites will probably want to have Google Analytics tracking, it seems like something worth mentioning in the VP docs.

### DIFF
--- a/docs/guide/basic-config.md
+++ b/docs/guide/basic-config.md
@@ -41,3 +41,16 @@ export default ({
   // ...apply enhancements to the app
 }
 ```
+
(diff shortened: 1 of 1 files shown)


ulivz:
Thanks, but we have supported for GA: https://vuepress.vuejs.org/config/#ga

brandonkelly:
Doh, I missed that. Maybe update the heading to “ga (Google Analytics)” so it shows up in the search results? (Or use Algolia DocSearch, which indexes the body contents too.)

codeofsumit:
how can I access the analytics object to send custom events and the likes from components? `this.$ga` is undefined

Issue #453, opened 2018-05-16
Need service worker update alert window.


EYHN:
In the default theme, when the service worker is updated, the page will not have any prompts. I suggest adding a alert window to remind the user that the page has been updated.

![update-toast](https://user-images.githubusercontent.com/13579374/40092524-47eb7628-58f0-11e8-84d7-40c841466029.png)

ulivz:
PR welcome. but the best to build a suitable vue-styled UI. ❤️

PR #462, opened 2018-05-17
feat: e.skipWaiting() in sw-update event

mysticatea:
Refs #453.

This PR adds the `e.skipWaiting()` method into the `sw-update` event in order to make that custom themes can implement reload buttons on the `sw-update` event.

**Background:**

I'm considering to use Vuepress to make the documentation of my OSSs. I will write the links to the documentation from changelogs. In the situation, Vuepress (with service workers) has a critical problem.

1. Browsers show the documentation from service worker caches when people open the documentation from the link in changelogs. It's old.
2. Reload doesn't work until all clients of the old service worker are closed.

So people have to open the documentation and close it and open it again in order to see the latest documentation. It's inconvenient.

Now Vuepress provides the `sw-update` event. It's the best timing to show "refresh" button to visitors in order to reload contents. However, new service worker doesn't available until they close all clients of the old service worker. Browsers will load contents from the old service worker even if it calls `location.reload(true)`. So I couldn't implement "refresh" button.

**Details:**

This PR adds the `e.skipWaiting()` method into the `sw-update` event to resolve that problem. The `e.skipWaiting()` method activates new service worker immediately. So we can reload contents after the `e.skipWaiting()` method is called. 

For example, [eslint-utils:/docs/.vuepress/theme/Layout.vue](https://github.com/mysticatea/eslint-utils/blob/89c9a6c146df0da1706e6f72919425a50e1618b8/docs/.vuepress/theme/Layout.vue) is an implementation of "refresh" button. If people clicked the "refresh" button, it calls the `e.skipWaiting()` method to available the new service worker then calls `location.reload(true)`.

**Notes:**

- This PR didn't implement the "refresh" button into the default theme because I don't have good idea about the way to configure the popup with internationalization.
- I'm not familiar with workbox. The `generateSW` method doesn't look to have the option which appends additional code. So I modified `build.js` to append a code to `service-worker.js` after workbox generated it. I'm not sure if this is good.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -81,11 +81,16 @@ module.exports = async function build (sourceDir, cliOptions = {}) {
   if (options.siteConfig.serviceWorker) {
     logger.wait('\nGenerating service worker...')
     const wbb = require('workbox-build')
-    wbb.generateSW({
+    await wbb.generateSW({
       swDest: path.resolve(outDir, 'service-worker.js'),
       globDirectory: outDir,
       globPatterns: ['**\/*.{js,css,html,png,jpg,jpeg,gif,svg,woff,woff2,eot,ttf,otf}']
     })
+    await fs.writeFile(
+      path.resolve(outDir, 'service-worker.js'),
+      await fs.readFile(path.resolve(__dirname, 'service-worker/skip-waiting.js'), 'utf8'),
+      { flag: 'a' }
+    )
   }
 
   // DONE.
--- a/lib/app/SWUpdateEvent.js
+++ b/lib/app/SWUpdateEvent.js
@@ -8,6 +8,13 @@ export default class SWUpdateEvent {
   }
 
   /**
+   * Check if the new service worker exists or not.
+   */
+  update () {
+    return this.registration.update()
+  }
+
+  /**
    * Activate new service worker to work 'location.reload()' with new data.
    */
   skipWaiting () {
--- a/lib/app/SWUpdateEvent.js
+++ b/lib/app/SWUpdateEvent.js
@@ -1,0 +1,35 @@
+export default class SWUpdateEvent {
+  constructor (registration) {
+    Object.defineProperty(this, 'registration', {
+      value: registration,
+      configurable: true,
+      writable: true
+    })
+  }
+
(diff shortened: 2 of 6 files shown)


EYHN:
Thanks for helping!

Sudden page refresh may interrupt readers' reading.
We may need a alert window.

mysticatea:
This PR adds the API to implement refresh UI.

lib/build.js
```diff
@@ -80,11 +80,16 @@ module.exports = async function build (sourceDir, cliOptions = {}) {
   if (options.siteConfig.serviceWorker) {
     console.log('Generating service worker...')
     const wbb = require('workbox-build')
-    wbb.generateSW({
+    await wbb.generateSW({
       swDest: path.resolve(outDir, 'service-worker.js'),
       globDirectory: outDir,
       globPatterns: ['**\/*.{js,css,html,png,jpg,jpeg,gif,svg,woff,woff2,eot,ttf,otf}']
     })
+    await fs.writeFile(
```
EYHN: Can we use the importScripts option instead?

https://developers.google.com/web/tools/workbox/modules/workbox-build#full_generatesw_config

mysticatea: Thank you for the suggestion.

I compared it and the current code, I didn't feel that the option improves the build steps much. We need the following steps with the option:

1. Copy the `service-worker/*.js` files to `outDir` manually. (Webpack doesn't handle the files for service worker because Workbox requires the files Webpack generated.)
2. Give the path to `service-worker/*.js` files to the `importScripts` option.

Step 1 is the same step as the current appending step essentially. Plus, there are some concerns:

1. the `importScripts` option makes extra HTTPS requests.
2. Chrome has a bug that overlooks the changes of imported scripts ([chromium#648295](https://bugs.chromium.org/p/chromium/issues/detail?id=648295)). This implies the change of the `service-worker/*.js` files will not trigger to update service worker's caches.

mysticatea:
I updated this PR for `register-service-worker@1.3.0` which probides service worker registration as a parameter.

ulivz:
Thanks for the great work again, let's leverage #533 !

PR #538, opened 2018-06-01
feat: Add support to import files as code fence

znck:
Add rule to import code snippets:

``` md
...
## Minimal

rollup-plugin-vue ships as zero config solution to package `.vue` files.

<<< @/cookbook/minimal/rollup.config.js{1,10}
```

screenshot-2018-6-2 rollup plugin for vue cookbook

### DIFF
--- a/lib/markdown/snippet.js
+++ b/lib/markdown/snippet.js
@@ -1,0 +1,43 @@
+const fs = require('fs')
+
+module.exports = function codeFrame(md, options = {}) {
+  const root = options.root || process.cwd()
+  function parser(state, startLine, endLine, silent) {
+    const CH = '<'.charCodeAt(0)
+    const pos = state.bMarks[startLine] + state.tShift[startLine]
+    const max = state.eMarks[startLine]
+
+    // if it's indented more than 3 spaces, it should be a code block
+    if (state.sCount[startLine] - state.blkIndent >= 4) {
+      return false
+    }
+
+    for (let i = 0; i < 3; ++i) {
+      const ch = state.src.charCodeAt(pos + i)
+      if (ch !== CH || pos + i >= max) return false
+    }
+
+    if (silent) {
+      return true
+    }
+
+    const start = pos + 3
+    const end = state.skipSpacesBack(max, pos)
+    const rawPath = state.src.slice(start, end).trim().replace(/^@/, root)
+    const filename = rawPath.split(/[{:\s]/).shift()
+    const content = fs.existsSync(filename) ? fs.readFileSync(filename).toString() : 'Not found: ' + filename
+    const meta = rawPath.replace(filename, '')
+
+    state.line = startLine + 1;
+
+    token = state.push('fence', 'code', 0)
+    token.info = filename.split('.').pop() + meta
+    token.content = content
+    token.markup = '```'
+    token.map = [startLine, startLine + 1]
+
+    return true
+  }
+
+  md.block.ruler.before('fence', 'code-frame', parser)
+}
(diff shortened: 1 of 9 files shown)


lib/markdown/snippet.js
```diff
@@ -0,0 +1,43 @@
+const fs = require('fs')
+
+module.exports = function codeFrame(md, options = {}) {
```
Akryum: `snippet`?

lib/markdown/snippet.js
```diff
@@ -0,0 +1,43 @@
...
+    token = state.push('fence', 'code', 0)
+    token.info = filename.split('.').pop() + meta
+    token.content = content
+    token.markup = '```'
+    token.map = [startLine, startLine + 1]
+
+    return true
+  }
+
+  md.block.ruler.before('fence', 'code-frame', parser)
```
ulivz: same here

ulivz:
Final thing, could you also document for this feature? thanks!

znck:
I feel it should support importing given lines from large a file but that can be a different PR.

ulivz:
Awesome! welcome to open any PRs for good stuffs like this.

octref:
I wish we could do a syntax more inline with markdown and Vue loaders, for example:
```markdown
# My Config

<code src="@/cookbook/minimal/rollup.config.js"></code>
<code src="@/cookbook/minimal/rollup.config.js" start="1" end="10"></code>
```

This is both valid markdown, and benefits include:

- Editor can support syntax highlight easily
- Editor might be able to linkify the `src` and offer jump-to-definition
- The code is now **semantic**. `<<<` is much less straight-forward and readable than `<code>`
- Following spec is always a good thing.

octref:
@ulivz That's actually quite a nice idea. I was about to suggest https://github.com/iainc/Markdown-Content-Blocks which would solve the problem of file transcultion in general.

One problem is `code` there actually don't mean anything. The other problem (also with the original proposal) is that `{1-2}` or `{1, 10}` is very misleading...I thought it was open/end but it was actually lines that need highlight. But I don't have a good idea now either.

znck:
@octref's concerns are valid. Let's give this problem another shot. 

What do we need:

- Transclusion (full or partial)
- Line highlighting
- Performed statically & within scope of markdown loader

@yyx990803 Any suggestions here?

----

> https://talk.commonmark.org/t/transclusion-or-including-sub-documents-for-reuse/270 
> (It's listed on [proposed commonmark extensions](https://github.com/commonmark/CommonMark/wiki/Proposed-Extensions))

znck:
Something like https://github.com/posva/markdown-it-custom-block could be better here.

``` md
:[code](../somefile.js)
:[code {1,4}](../somefile.js)
:[code js{1,4}](../somefile.js)
```
It would support explicit language, line highlighting and with click jumping editor support.

ulivz:
@znck I like it. but how about this syntax for `partial transclusion`?

```md
@[code{1-10} js{1,4}](../somefile.js)
```

It prevent the links from being changed and still offer jump-to-definition.

znck:
I'll send a PR by EOD.

octref:
I think the highlight line syntax need more thought. Can you do things like `{2-5,8-10}` for ranges?
On the other hand I wouldn’t find partial transclution very useful or ergonomic. When I go to the transcluded file I don’t know which part is included without looking at the importing file. Also if I only want line 2-5 why don’t I delete line 1? If Line 1 is necessary why not just highlight line 2-5?

znck:
If you have a large file and you want to import only couple of lines. Example importing only script from SFC.

octref:
Well you can also remove other lines in the source file, if they are not relevant.
Maybe you are thinking in one project, you can import in `/doc/index.md` something from `/src/Component.vue`, etc. But those are very implicit dependencies and I don't think that's a real need.

daggerok:
Hey, so can anybody tell me if now we can include partially code snippets or it's still not possible?

PR #556, opened 2018-06-07
feat: better sidebar transition

ulivz:
# Summary

This PR leverages the transition of [vue.js official website](https://vuejs.org/v2/guide/) and also enhances the mobile's style when sidebar slide in from the left side. **It hides header, which allows users to focus more on sidebar.**

### Before:
option-0

### Option 1
option-1

Hmmm... the `box-shadow` looks not very good in this gif, this `box-shadow` is also migrated from official website. so add a extra static snapshot here:

image

### Option 2
option-2

@yyx990803 which one do you prefer?

### DIFF
--- a/lib/default-theme/Sidebar.vue
+++ b/lib/default-theme/Sidebar.vue
@@ -1,6 +1,5 @@
 <template>
   <div class="sidebar">
-    <slot name="header"/>
     <NavLinks/>
     <slot name="top"/>
     <ul class="sidebar-links" v-if="items.length">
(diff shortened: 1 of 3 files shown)


AWulkan:
I don't like that it hides the hamburger menu. I think it's intuitive for a user to click the hamburger menu one more time to close the side bar.

AWulkan:
But it's there any real benefit of hiding it? I don't see how it makes the UX better. Plus, if you accidentally double click you will end up at the link that popped up in the hamburger menues previous position.

Anyway, just giving my opinion.

I like the ability to swipe/tap outside to close though, but i don't think that they need to replace the menu button. It makes the UI inconsistent. Also, how does it work on a desktop where the window might be small enough to trigger the mobile view? Then it is unintuitive to not have a button imo.

PR #579, opened 2018-06-13
allow markdown.config in config.yml fix #326

### LINKED ISSUE #326: Possible to extend markdown while using config.yml?
```
title: Rules
description: Short references
markdown:
    config: md => { md.use(require('markdown-it-deflist')) }
```

When written like this, the markdown config is turned into a string and not evaluated as a function. Is this intentional?

neumayr:
* allow setup of `markdown.config` in `config.yml`

`markdown.config` is a `String` and not evaluated as a function if you use `config.yml`

```yaml
# .vuepress/config.yml
title: Awesome docs
description: Vuepress rocks
markdown:
  lineNumbers: true
  config: |
    md => {
      md.use(require('markdown-it-checkbox'))
      md.use(require('markdown-it-ins'))
      md.use(require('markdown-it-mark'))
    }
```

:link: https://github.com/vuejs/vuepress/issues/326

/cc @luminarious

---

**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

**Other information:**

We aren't ecmascript pros. And we know `eval` is evil in some context. But we had no better idea…

### DIFF
--- a/lib/markdown/index.js
+++ b/lib/markdown/index.js
@@ -50,6 +50,9 @@ module.exports = ({ markdown = {}} = {}) => {
 
   // apply user config
   if (markdown.config) {
+    if (typeof markdown.config === 'string') {
+      markdown.config = eval(markdown.config)
+    }
     markdown.config(md)
   }
 


ulivz:
Thanks for your work. but we don't think we should support for that.

As you can imagine, when the user does not give `config` a valid JavaScript string, this will give an error.

**So if you need anything JavaScript, please use JS. thanks.**

neumayr:
@ulivz Is there a way to do further checks about this. For example, if the config format is `.yml`. 
Imho, reporting an error is 10x better than not supporting this feature for a supported config format.

Another idea is to allow a `config.js` and `config.yml` get merged. I could life with only doing the `markdown.config` in the JS file.

Issue #610, opened 2018-06-23
Option to disable svg in external links


octref:
`<OutboundLink>` generates an icon like this:

![image](https://user-images.githubusercontent.com/4033249/41814347-2bd0d3d0-76fe-11e8-9230-70e63e3fef18.png)

There should be an option to turn this off.

octref:
It make sense to include it in the theme, however including it in the default markdown renderer and not offering anyway to turn it off is forcing this icon to everyone, even those who's not using the default theme.

dcastil:
What is the status of this feature? I would like to be able to disable the OutboundLink component and think it should be rather opt-in with a plugin.

xinconan:
How to disable svg in external links? The web site has  removed svg, https://vuepress.vuejs.org/config/, but I can't find config option.

ulivz:
See: https://github.com/vuejs/vuepress/pull/614#issuecomment-472482023

morphatic:
If you only want to hide the outbound link icon only in certain cases you can hide it with CSS. For example, in my site I sometimes wrap images in external links, e.g.

```markdown
[![some img](/assets/images/some.png)](http://someothersite.com)
```

The HTML rendered by this looks something like:

```html
<a href="http://someothersite.com" target="_blank" rel="noopener noreferrer">
  <img src="/assets/images/some.png" alt="some img">
  <svg class="icon outbound">... this is the external link icon ...</svg>
</a>
```

And results in an image with the external link icon after it like this:

![some img](https://imgur.com/gL4xXvU)

In my CSS, I can use a sibling selector to hide the icon like this:

```css
img + .icon.outbound {
  display: none;
}
```

This is pretty flexible and wouldn't require you to disable all of `markdown-it`. HTH!

deancn:
```
  markdown: {
    externalLinkSymbol: false,
  },
```

![image](https://user-images.githubusercontent.com/1617384/76280663-8bb60100-62cd-11ea-826e-67a853969aed.png)


do not work yet.  Any thing wrong?

PR #614, opened 2018-06-26
feat: option to disable svg in external links (close: #610)

ulivz:
**Summary**

- Close: #610 
- Release at: **0.11.0**

cc @octref 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

**Other information:**

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -163,6 +163,17 @@ Function for transforming header texts into slugs. This affects the ids/links ge
 
 The key and value pair will be added to `<a>` tags that points to an external link. The default option will open external links in a new window.
 
+### markdown.externalLinkSymbol
+
+- Type: `boolean`
+- Default: `true`
+
+By default (<OutboundLink/>) will be followed by each external link, you can disable it via setting this option to `false`.
+
+**Also see:**
+
+- [Built-In Components > OutboundLink](../guide/using-vue.md#outboundlink) 
+
 ### markdown.anchor
 
 - Type: `Object`
--- a/docs/guide/using-vue.md
+++ b/docs/guide/using-vue.md
@@ -180,7 +180,11 @@ export default {
 
 ### OutboundLink <Badge text="stable"/>
 
-It(<OutboundLink/>) is used to indicate that this is an external link. In VuePress this component have been followed by every external link.
+It(<OutboundLink/>) is used to indicate that this is an external link. In VuePress this component have been followed by each external link.
+
+**Also see:**
+
+- [Config > markdown.externalLinkSymbol](../config/README.md#markdown-externallinksymbol) 
 
 ### ClientOnly <Badge text="stable"/>
 
--- a/docs/zh/config/README.md
+++ b/docs/zh/config/README.md
@@ -162,6 +162,17 @@ module.exports = {
 
 这个键值对将会作为特性被增加到是外部链接的 `<a>` 标签上，默认的选项将会在新窗口中打开一个该外部链接。
 
+### markdown.externalLinkSymbol
+
+- Type: `boolean`
+- Default: `true`
+
+默认情况下，(<OutboundLink/>) 会紧跟在每一个外部链接后面，你可以通过将这个选项配置成 `false` 来禁用。
+
+**Also see:**
+
+- [内置的组件 > OutboundLink](../guide/using-vue.md#outboundlink) 
+
 ### markdown.toc
 
 - 类型: `Object`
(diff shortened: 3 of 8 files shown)


octref:
I feel it might be better to put the option inside `externalLinks`, something like
```
externalLinks: {
  attributes: { target: '_blank', rel: 'noopener noreferrer' },
  symbol (or call it icon?): true | false
}
```

However, that is backward incompatible...However I feel it's better to have options organized and deprecate the original usage, now Vuepress is still pre 1.0.

ulivz:
Agree with that, let's reorganize it. 

lol, I'm not very good at name, `icon` looks better to me.

octref:
Sorry for all the opinions - but I feel this should actually be an option contributed from the theme with your plugin API. I think it would be great to start that early, so there can be clean separation between theme / core config.

ulivz:
Actually, plugin API is almost done but I still have to spend some time optimizing it, and I also agree with that anything that can be done by plugin should be done by plugin.

But users are unable to modify the implementation of built in `makrdown-it` plugins through VuePress's plugin API. especially for `link`, as a core plug-in, we may not be able to separate it from core.

IMO, we can implement a low-level API at core for that, and each theme can determine its `public API / preset` by using plugin APIs.

octref:
@ulivz Thanks - to keep the markdown clean, I think some common functionalities people expect from a MD renderer such as anchor and slug should be built-in, but also possible to turn off.
For everything cosmetic, they should go to the theme.

destroytoday:
Hi! What's remaining with this PR in order to get it merged? (I'm eager to get it in since I just hit this issue myself when making a custom theme).

If the config var simply needs to change, I'd be happy to make a contribution to get everything wrapped up. Was something like the following the consensus? (to keep backwards-compatibility)

```
externalLinks: {
  target: '_blank',
  rel: 'noopener noreferrer',
  icon: true | false
}
```

(I'm not entirely sure if the most recent comments about using the plugin API were the final call, but for what it's worth, before finding the issue/PR, I expected to be able to set a boolean in the config)

zlliang:
Also eager to get it!
For now I disable the icon through CSS:
```css
svg.icon.outbound {
  display: none;
}
```

robmuh:
If you have ejected the default theme with 0.14.4 or so you can put the following in the `.vuepress/theme/styles/theme.style` file:

```stylus
.icon.outbound
  display none !important 
```

ulivz:
In 1.x we can acquire it by removing the built-in markdown-it [link](https://github.com/vuejs/vuepress/blob/master/packages/%40vuepress/markdown/lib/link.js) plugin and creating a custom link plugin:

```js
// .vuepress/config.js
module.exports = {
  chainMarkdown (config) {
    const { PLUGINS } = require('@vuepress/markdown')

    config
      .plugins
        .delete(PLUGINS.CONVERT_ROUTER_LINK)

    config
      .plugin(PLUGINS.CONVERT_ROUTER_LINK)
        .use(/* Your custom link plugin */)
  }
}
```

adam-lynch:
For anyone else who needs it, this uses the same link plugin (so it works the exact same and supports the same options) except `<OutboundLInk/>` is removed from the result.

```js
// .vuepress/config.js
module.exports = {
  chainMarkdown(config) {
    const { PLUGINS } = require('@vuepress/markdown')
    const originalLinkPlugin = require('@vuepress/markdown/lib/link.js');

    config
      .plugins
        .delete(PLUGINS.CONVERT_ROUTER_LINK)

    const linkPlugin = function (md) {
      const result = originalLinkPlugin.apply(this, arguments);
      const close = md.renderer.rules.link_close;
      md.renderer.rules.link_close = function() {
        return close.apply(this, arguments).replace('<OutboundLink/>', '');
      }
      return result;
    };

    config
      .plugin(PLUGINS.CONVERT_ROUTER_LINK)
        .use(linkPlugin, [{
          // The config.markdown.externalLinks options https://vuepress.vuejs.org/config/#markdown-externallinks
          target: '_blank',
          rel: 'noopener noreferrer'
        }])
  },
};
```

deancn:
@ulivz  is this update to docs? 
https://vuepress.vuejs.org/config/#markdown

steveellis:
Removing the external link icon is tricky but doable with CSS. What caused me some pain was that the HTML generated from my build server was slightly different than the one locally, requiring different selection rules to excise it.

This is what can work when it is built locally.... It doesn't have the span, it's just an svg inside the external link (here selected with the target attribute). There's probably a more precise selector but whatever...

```
a[target] svg {
  display: none !important;
}
```

This is what it looks like when built with yarn in the development pipeline.
```
a[target] span { 
  display: none !important
} 
```
Just putting both those rules in theme.styl worked for me.

PR #655, opened 2018-07-17
fix: exclude dest folder from pageFiles (fixes #654)

### LINKED ISSUE #654: Build fails if dest folder contains a README.md file
#### Version
0.12.0
#### Steps to reproduce
Change the dest folder to something else on the root on the project.
```
module.exports = {
    dest: 'dist',
};
```
Add a READE.md file to `.vuepress/public` and launch the build task.
The first time everything will work just fine, you will have the new dest folder with all the files.
Launch the build task again and you will see a *Error: Failed to compile with errors.* error on the console.

[Reproduction repo](https://github.com/pedrodim/vuepress-bug)

#### What is expected?
Build to complete with no errors just as the first time it was launched, when no dest folder with the README.md existed on the project's root.

Probably vuepress is trying to parse the dest folder as a normal src folder instead of excluding it from parsing.

Also, the dest folder gets deleted after the script fails, maybe it should delete dest before parsing the files to compile?

- Your OS: macOS 10.13.6
- Node.js version: v8.11.3
- Browser version: Chrome 67.0.3396.99
- Is this a global or local install? Local install
- Which package manager did you use for the install? NPM

pedrodim:
**Summary**
Fixes #654

Dest folder, if present in the config, gets excluded from pageFiles to prevent parsing errors if any .md file's present.

- [x] Bugfix

- [x] No

### DIFF
--- a/lib/prepare/resolveOptions.js
+++ b/lib/prepare/resolveOptions.js
@@ -106,6 +106,12 @@ module.exports = async function resolveOptions (sourceDir) {
   const markdown = createMarkdown(siteConfig)
 
   // resolve pageFiles
+  const files = ['**/*.md', '!.vuepress', '!node_modules']
+  if (siteConfig.dest) {
+    // exclude dest folder from pageFiles
+    const outDirRelative = path.relative(sourceDir, outDir)
+    files.push('!' + siteConfig.dest)
+  }
   const pageFiles = sort(await globby(['**/*.md', '!.vuepress', '!node_modules'], { cwd: sourceDir }))
 
   // resolve lastUpdated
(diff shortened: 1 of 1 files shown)


lib/prepare/resolveOptions.js
```diff
@@ -106,6 +106,12 @@ module.exports = async function resolveOptions (sourceDir) {
   const markdown = createMarkdown(siteConfig)
 
   // resolve pageFiles
+  const files = ['**/*.md', '!.vuepress', '!node_modules']
+  if (siteConfig.dest) {
+    // exclude dest folder from pageFiles
+    const outDirRelative = path.relative(sourceDir, outDir)
+    files.push('!' + siteConfig.dest)
```
ulivz: You just created a new variable, but where do you use it?

pedrodim: Sorry about that, fixed now.

PR #689, opened 2018-07-31
fix: strip html from meta title, if present; fix #688

### LINKED ISSUE #688: Badge Component HTML displays in Page Title when used in H1
#### Version
0.13.0

#### Steps to reproduce
1. clone [this repository](https://github.com/roberthamel/vuepress-badge-issue)
2. run `vuepress dev`

#### What is expected?

Ideally, I would think that the following would occur: 
- the text `SAMPLE: Sample Page Title` would be parsed into the head title
- the text `SAMPLE: Sample Page Title` would be parsed into the sidebar

At the very least, the HTML should not display.

- The head title displays `<Badge text="SAMPLE"/> Sample Page Title`.
- The sidebar displays `<Badge text="SAMPLE"/> Sample Page Title`.

screen shot 2018-07-31 at 11 06 44 am

I am aware that I can override the header using the following front matter:

```yml
---
title: Sample Page Title
---
```

But, I have a situation where I want to use the badge component to differentiate the page from others.

I am able to use regex to override the theme in order to fix the sidebar, but from what I can tell, the page head title is generated by the vuepress cli.

- Your OS: **MacOS**
- Node.js version: **10.7.0**
- Browser version: **Chrome 67**
- Is this a global or local install? **global** 
- Which package manager did you use for the install? **yarn**

roberthamel:
closes #688

**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**
When the Badge component is used in an `h1` tag that will be used as the site's meta title, the html will no longer display.

### DIFF
--- a/lib/app/root-mixins/updateMeta.js
+++ b/lib/app/root-mixins/updateMeta.js
@@ -16,7 +16,6 @@ export default {
         ? this.$title.replace(/^(\<.*\/\>)/, '')
         : this.$title
       document.title = title
-      document.title = this.$title
       document.documentElement.lang = this.$lang
       const userMeta = this.$page.frontmatter.meta || []
       const meta = userMeta.slice(0)
--- a/lib/app/root-mixins/updateMeta.js
+++ b/lib/app/root-mixins/updateMeta.js
@@ -12,6 +12,10 @@ export default {
     this.currentMetaTags = new Set()
 
     const updateMeta = () => {
+      const title = this.$title.match(/\<.*\/\>/)
+        ? this.$title.replace(/^(\<.*\/\>)/, '')
+        : this.$title
+      document.title = title
       document.title = this.$title
       document.documentElement.lang = this.$lang
       const userMeta = this.$page.frontmatter.meta || []


ulivz:
Cool~

This is my usage:

1. Usage

``` md
# Introduction <Badge text="beta" type="warn"/>
```

2. UI

image

3. Meta snapshot

image

Since I cannot reproduce it. could you provide more details as above. thanks.

roberthamel:
The scenario I am using is this:

```
# <Badge text="beta" type="warn"/> Introduction
```

I am unclear as to why it was designed to only strip from the end. My understanding is that the title of the document should not have any markup in it. If the title were to specifically need it, it could be described in different ways without the markup.

roberthamel:
Looking through other Vue documentation sites, I am seeing some places where markup is used in a heading.

For example, [this page](https://router.vuejs.org/api) contains a couple of headers containing markup. 

In this case, having a badge before the header text would not be ideal for the theme. Since I can eject the theme and override it, that's what I chose to do. The document title is internal to the vuepress-cli.

This pr would not affect the document head for that specific page since all markdown headings are not contained in h1 tags.

I will continue to research other scenarios where this pr would not be ideal.

ulivz:
@roberthamel Let me reply to your question one by one:

1. **Q**: I am unclear as to why it was designed to only strip from the end.

   **A**: Since the initial user case is `feature name + status`, so it's intended to only strip from the end. **But we can also support strip from the left.**, the reason is in `2`.

2. Q: Why does markup in other Vue docs docs work?

   There are two scenarios:

  ``` js
  // #1 Doesn't have "`" wrapped, output will be: `b`
  <a> b

  // #2 Have "`" wrapped, output will be: `<cpde>a<code> b`
  `<a>` b
  ```
  For 1,  the markup will be considered as the real HTML, we don't know its output at compile time, so it should be stripped from the header, while in 2, the markup was wrapped in a inline code identifiers, so it will be considered as the header's content.

 And this PR wasn't a real fix, I will work for `strip from the left` by myself. If you have interest, you can check out our [test case](https://github.com/vuejs/vuepress/blob/master/test/util/parseHeaders.spec.js) for parsing headers, and also our [changelog](https://github.com/vuejs/vuepress/blob/master/CHANGELOG.md).

roberthamel:
I have closed the issue related to this. I like your explanation and direction with this. I would like to help make this happen. I definitely need more practice understanding the depth of it. I can close this pull request, if you want.

Issue #823, opened 2018-09-11
Scroll overflow triggered by long words


FistMeNaruto:
#### Version
0.14.2
#### Steps to reproduce

1. Open any VuePress project (https://vuepress.vuejs.org/guide/)
2. Open devtools and edit in a long word inside the content (a URL like `https://github.com/vuejs/vuepress/issues/vuepress/issues/vuepress/issues/`, for example)

#### What is expected?

Long word breaking and scrolling working properly:

![screenshot from 2018-09-11 23-35-46](https://user-images.githubusercontent.com/13431677/45386272-9aa9ed80-b61b-11e8-985b-78e12f8309e7.png)

#### What is actually happening?

Long word introduces a scroll overflow, which allows the navbar to be scrolled out of the viewport:

![screenshot from 2018-09-11 23-36-17](https://user-images.githubusercontent.com/13431677/45386316-bf9e6080-b61b-11e8-8b43-3b49ad189c85.png)

The same issue happens when long words are inside the nav menu:

![screenshot from 2018-09-11 23-11-17](https://user-images.githubusercontent.com/13431677/45386563-510dd280-b61c-11e8-9406-fd8b60bf3577.png)

Or section headers:

![screenshot from 2018-09-11 23-42-47](https://user-images.githubusercontent.com/13431677/45386607-6d117400-b61c-11e8-9f37-8d5b472a9789.png)

#### Other relevant information

- Your OS: Arch linux
- Node.js version: v10.10.0
- Browser version: Chromium 68.0.3440.106
- Is this a global or local install? Local
- Which package manager did you use for the install? Yarn

ulivz:
See: https://github.com/vuejs/vuepress/pull/824#pullrequestreview-155719501

PR #824, opened 2018-09-11
fix: #823

FistMeNaruto:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

Before:
screenshot from 2018-09-11 23-36-17
After:
screenshot from 2018-09-11 23-35-46

Before:
screenshot from 2018-09-11 23-11-17
After:
screenshot from 2018-09-11 23-11-06

Before:
screenshot from 2018-09-11 23-42-47
After:
screenshot from 2018-09-11 23-54-03

- [ ] Yes
- [x] No (hopefully)

- [x] Chrome
- [x] Firefox (long words do not break nav bar, but still introduce overflow-x; also `word-break: break-word;` does not work on Firefox, therefore menu links still oveflow, but overflow is hidden) 
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

Some information about the css properties used:
https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/

### DIFF
--- a/lib/default-theme/styles/theme.styl
+++ b/lib/default-theme/styles/theme.styl
@@ -16,6 +16,9 @@ body
   -moz-osx-font-smoothing grayscale
   font-size 16px
   color $textColor
+  overflow-wrap break-word
+  word-wrap break-word
+  word-break break-word
 
 .page
   padding-left $sidebarWidth


ekoeryanto:
I think we should use the specific tags selector instead of `body`

FistMeNaruto:
Generally I would agree with that, but in this case `body` targets both the content and menu. I don't see any case where long word breaking would introduce unwanted effects. But yes, specific tags can be used here aswell.

FistMeNaruto:
I have brought this up mainly because I had an issue with a long URL in my own project (see first two pictures) and it introduced scroll overflow. I think it is relatively common to have URLs in text and it does make sense to break them in case they overflow.

Issue #966, opened 2018-10-29
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory


quanzaiyu:
## Bug report

When running `vuepress build website`, it reports an error

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

#### Version

1.0.0-alpha.17

#### Steps to reproduce

`package.json`

```json
{
  "scripts": {
    "build": "vuepress build website"
   }
}
```

run `yarn build`

#### What is expected?

Fix this error.

#### What is actually happening?

```
(node:27516) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
[22:46:54] Compiled Server in 27s
[22:47:02] Compiled Client in 34s
 WAIT  Rendering static HTML...
Rendering page: /books/road_vue_router/Guard.html
<--- Last few GCs --->

[27516:000001E86E570900]   244905 ms: Scavenge 1318.3 (1423.9) -> 1317.6 (1423.9) MB, 2.8 / 0.0 ms  (average mu = 0.112, current mu = 0.062) allocation failure
[27516:000001E86E570900]   244923 ms: Scavenge 1318.6 (1423.9) -> 1318.1 (1423.9) MB, 3.7 / 0.0 ms  (average mu = 0.112, current mu = 0.062) allocation failure
[27516:000001E86E570900]   244930 ms: Scavenge 1318.6 (1423.9) -> 1318.2 (1424.4) MB, 2.8 / 0.0 ms  (average mu = 0.112, current mu = 0.062) allocation failure


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0000030D49F5C6C1]
    1: StubFrame [pc: 0000030D49F4E2EC]
Security context: 0x0268e2c1e681 <JSObject>
    2: replace [00000268E2C10C01](this=0x02e8c72ae589 <String[1]: />,0x006b1bec5111 <JSRegExp <String[25]: ([.+*?=^!:${}()[\]|\/\\])>>,0x008fedf7a139 <String[3]: \$1>)
    3: addRouteRecord(aka addRouteRecord) [0000020FFDFAEBB9] [D:\Workplace\__Blog__\quanzaiyu\node_modules\vue-router\dist\vue-router.common.js:~1137]...

[...]

PR #1068, opened 2018-12-04
perf: improve ssr performance

janvennemann:
**Summary**

Improve ssr performance by caching items when resolving sidebar items. Fix memory leak in root mixins.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [x] Other, please describe:

We are testing VuePress on a huge API documentation which currently has over 400 pages. 200 of those pages are reachable via the sidebar in different sidebar groups. We experienced very slow build times (ranging from 300 - 800ms per page) and crashes due to heap out of memory.

After a little investigation we identified that resolving of sidebar items is responsible for slow build times. This introduces a simple map that stores page objects for each normalized path for faster lookups. The build times are now back to a reasonable 50-100ms for each page.

The crash was caused by the `updateLoadingState` mixin, which apparently contains client only code that does not make much sense to run on the server. The event handler bindings in that mixin will lead to a memory leak since the components will never be released.

- [ ] Yes
- [x] No

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**
Tested on Node v10.13.0

#966 may be related.

For a reproducible test case, clone https://github.com/appcelerator/titanium-vuepress-docs and build with `npm run docs:build`


packages/@vuepress/core/lib/app/root-mixins/updateLoadingState.js
```diff
@@ -2,6 +2,10 @@ import SmoothScroll from 'smooth-scroll/dist/smooth-scroll.js'
 
 export default {
   created () {
+    if (this.$ssrContext) {
```
ulivz: Actually this file has been removed for now since we deprecated smooth scroll.

packages/@vuepress/theme-default/util/index.js
```diff
@@ -128,12 +126,14 @@ export function resolveSidebarItems (page, regularPath, site, localePath) {
   }
 
   const sidebarConfig = localeConfig.sidebar || themeConfig.sidebar
+  let normalizedPagesMap = {}
+  pages.forEach(page => normalizedPagesMap[normalize(page.regularPath)] = page);
```
ulivz: Maybe following style would be more refreshing:

```js
const normalizedPagesMap = pages.reduce((map, page) => map[normalize(page.regularPath)] = page, {})
```

janvennemann:
@ulivz i'll update the PR tomorrow

janvennemann:
@ulivz fixed!

PR #1075, opened 2018-12-06
✨ Router scroll behaviour, fix #1071

### LINKED ISSUE #1071: Add native-like scroll behaviour when changing routes
## Feature request

Currently the user stays at his current scroll position when the route is being changed. But when the user gets to a new page, he should see the top of the new page and when he goes back- and forward he should see the scroll position at which he left the page previously.

The logic on how the scroll position changes based on a route change should look like this:

* New page → top scroll position
* Already visited page → scroll position at which user previously left the page
* Anchor change → scroll position at anchor

* The content of a page is ordered from top to bottom, so the top is what the user should see first when navigating to a page.
* The user expects to be at the top of a page when the route changes, everything else can lead to confusion.

The VuePress API doesn't change with this feature.

Vue Router provides a [scrollBehavior API](https://router.vuejs.org/guide/advanced/scroll-behavior.html) which can set the scroll position when changing routes. This could be implemented into the router initialisation in `packages/@vuepress/core/lib/app/app.js`.

The anchor scrolling doesn't need to be implemented as this is already solved, but maybe it's better to put everything into one place.

Yes, just want to check if this is desired.

dcastil:
**Summary**

closes #1071 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

  → see #1071 

**Other information:**

I removed `smooth-scroll` as it was causing issues when used with `active-header-links`.

### DIFF
--- a/packages/@vuepress/core/lib/app/app.js
+++ b/packages/@vuepress/core/lib/app/app.js
@@ -23,7 +23,7 @@ if (module.hot) {
     if (siteData.base !== prevBase) {
       window.alert(
         `[vuepress] Site base has changed. ` +
-        `Please restart dev server to ensure correct asset paths.`
+          `Please restart dev server to ensure correct asset paths.`
       )
     }
   })
@@ -62,15 +62,25 @@ export function createApp (isServer) {
     base: siteData.base,
     mode: 'history',
     fallback: false,
-    routes
+    routes,
+    scrollBehavior (to, from, savedPosition) {
+      if (savedPosition) {
+        return savedPosition
+      }
+      if (to.path !== from.path) {
+        return { x: 0, y: 0 }
+      }
+    }
   })
 
   // redirect /foo to /foo/
   router.beforeEach((to, from, next) => {
     if (!/(\/|\.html)$/.test(to.path)) {
-      next(Object.assign({}, to, {
-        path: to.path + '/'
-      }))
+      next(
+        Object.assign({}, to, {
+          path: to.path + '/'
+        })
+      )
     } else {
       next()
     }
@@ -94,7 +104,11 @@ export function createApp (isServer) {
       render (h) {
         return h('div', { attrs: { id: 'app' }}, [
           h('router-view', { ref: 'layout' }),
-          h('div', { class: 'global-ui' }, globalUIComponents.map(component => h(component)))
+          h(
+            'div',
+            { class: 'global-ui' },
+            globalUIComponents.map(component => h(component))
+          )
         ])
       }
     })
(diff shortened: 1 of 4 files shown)


packages/@vuepress/core/lib/app/app.js
```diff
@@ -62,15 +62,25 @@ export function createApp (isServer) {
     base: siteData.base,
     mode: 'history',
     fallback: false,
-    routes
+    routes,
+    scrollBehavior (to, from, savedPosition) {
```
janvennemann: What about cases when `to.hash` is set?

packages/@vuepress/core/lib/app/app.js
```diff
@@ -62,15 +62,25 @@ export function createApp (isServer) {
...
+    }
   })
 
   // redirect /foo to /foo/
   router.beforeEach((to, from, next) => {
     if (!/(\/|\.html)$/.test(to.path)) {
-      next(Object.assign({}, to, {
-        path: to.path + '/'
-      }))
+      next(
```
ulivz: Do not format code and only focus what you are doing.

packages/@vuepress/core/lib/app/app.js
```diff
@@ -94,7 +104,11 @@ export function createApp (isServer) {
       render (h) {
         return h('div', { attrs: { id: 'app' }}, [
           h('router-view', { ref: 'layout' }),
-          h('div', { class: 'global-ui' }, globalUIComponents.map(component => h(component)))
+          h(
```
ulivz: Same here.

packages/@vuepress/core/lib/app/app.js
dcastil: That interferes with `plugin-active-header-links` unfortunately. It causes the website to jump around when the anchor is being pushed.

packages/@vuepress/core/lib/app/app.js
dcastil: Ah sorry, I'm using prettier by default.

packages/@vuepress/core/lib/app/app.js
janvennemann: Alright perfect, then i guess in the old implementation it was just to disable scroll behavior via some flag.

dcastil: Ah that would make sense here, too! Have to check that out later.

PR #1195, opened 2019-01-15
Sunsetting webpack-serve

ulivz:
> _TL;DR:_  Any technology has a suitable application scenario. For vuepress, `express` is actually more suitable than `koa`.

### Why do we abandon `webpack-serve` and choose `webpack-dev-server`?

There are several main reasons for us to embrace `webpack-dev-server`.

1. The underlying shelf that we used to implement `dev-server` and `HMR` is [webpack-serve](https://github.com/webpack-contrib/webpack-serve) which is based on [koa](https://github.com/koajs/koa), not long ago the maintainer announced the [deprecation](https://github.com/webpack-contrib/webpack-serve#webpack-serve) of `webpack-serve` and recommended us to choose [webpack-dev-server](https://github.com/webpack/webpack-dev-server/) that is based on [express](https://github.com/expressjs/express).

2. Due to its more succinct usage, `webpack-dev-server` is used by many popular build tools, e.g.  [vue-cli](https://github.com/vuejs/vue-cli), [poi](https://github.com/egoist/poi) etc. combining with following 3rd point, you'll find that `webpack-dev-server` will help us to reduce the cost of switching between similar projects and writing plugins.

3. In the 1.0 release plan, we supported a plugin API [enhanceDevServer](https://vuepress.vuejs.org/plugin/option-api.html#enhancedevserver) which allows us to tweak the instance of underlying koa app. although koa‘s nonlinear architectural model looks cooler, but the reality is that the ecology of express is better, as well as the cost of writing middleware is lower than koa. 

   e.g. in koa, if you want to use [connect-history-api-fallback](https://github.com/bripkens/connect-history-api-fallback), you'll need [koa-connect](https://github.com/vkurchatkin/koa-connect) for which you needn't care in express.

### Migration

**For 0.x users**, there is no any effect since we didn't expose API to modify it.
**For 1.x users whose version is lower than 1.0.0-alpha.32**:

   1. You should use `beforeDevServer`(e.g. [before](https://webpack.js.org/configuration/dev-server/#devserver-before) in webpack-dev-server) to replace `enhanceDevServer`;
   2. You can use `afterDevServer`(e.g. [after](https://webpack.js.org/configuration/dev-server/#devserver-after) in webpack-dev-server) to execute custom middleware after all other middleware internally within the server.

   e.g. writing a plugin to create sub public directory `images`:

   1. In the past (_**>= 1.0.0-alpha.0 && <= 1.0.0-alpha.32**_):

  ```js
  const { fs, path } = require('@vuepress/shared-utils')
  const mount = require('koa-mount')
  const serveStatic = require('koa-static')
  
  module.exports = (options, context) => {
    const imagesAssetsPath = path.resolve(context.sourceDir, '.vuepress/images')
  
    return {
        // For development
        enhanceDevServer (app) {
          app.use(mount(path.join(context.base, 'images'), serveStatic(imagesAssetsPath)))
        },
  
        // For production
        async generated () {
          await fs.copy(imagesAssetsPath, path.resolve(context.outDir, 'images'))
        }
    }
  }
  ```

   2. With this PR (_**>= 1.0.0-alpha.33**_):

  ```js
  const { fs, path } = require('@vuepress/shared-utils')
  const express = require('express')
  
  module.exports = (options, context) => {
    const imagesAssetsPath = path.resolve(context.sourceDir, '.vuepress/images')
  
    return {
      // For development
      beforeDevServer (app) {
        app.use('/images', express.static(imagesAssetsPath))
      },
  
      // For production
      async generated () {
        await fs.copy(imagesAssetsPath, path.resolve(context.outDir, 'images'))
      }
    }
  }
  ```

---

### What kind of change does this PR introduce? (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

- [x] Chrome
- [ ] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE


packages/@vuepress/core/lib/dev.js
```diff
@@ -105,52 +100,49 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
     config = applyUserWebpackConfig(userConfig, config, false /* isServer */)
   }
 
+  const serverConfig = {
+    disableHostCheck: true,
```
ulivz: When set to true this option bypasses host checking

packages/@vuepress/theme-blog/README.md
```diff
@@ -1,3 +1,3 @@
 # @vuepress/theme-blog
 
-> theme-blog for vuepress
\ No newline at end of file
+> theme-blog for vuepress (WIP)
```
meteorlxy: I think these blog related changes should not be included in this PR

packages/@vuepress/core/lib/dev.js
```diff
@@ -105,52 +100,49 @@ module.exports = async function dev (sourceDir, cliOptions = {}) {
...
-    port,
-    open: cliOptions.open,
-    add: app => {
-      // apply plugin options to extend dev server.
-      ctx.pluginAPI.options.enhanceDevServer.syncApply(app)
-
-      const userPublic = path.resolve(sourceDir, '.vuepress/public')
-
-      // enable range request
-      app.use(range)
```
ulivz: ✅Express enable range request in development. so removing it will not break the fix of https://github.com/vuejs/vuepress/pull/555.

meteorlxy:
陈哥你脾气别这么大嘛，我好好写了半天想和你讨论的嘛……

你说的 ready hook 是指的 vuepress 的 lifecycle 吗，这个和 router resolved 的关系我确实不太懂啊能不能指教一下，加个微信行不 😢

PR #1228, opened 2019-01-25
refactor($core): make LayoutDistributor functional

meteorlxy:
**Summary**

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

If changing the UI of default theme, please provide the **before/after** screenshot:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

**Other information:**

It's better to make this simple component functional, as it is transparent to users.

Issue #1226, opened 2019-01-24
SPA application layout wrapper needed


SGarno:
## Feature request
Please create an application level layout component that gets applied to all layouts.   This allows for a SPA type application to wrap each layout with a header/footer/sidebar etc without having to import components in the Layout.vue and then having conditional logic or rendering techniques for deciding which layout component to include.  

#### What problem does this feature solve?
Not having this feature requires coding in Layout.vue to identify what layout components are to be used for a particular page.   It also forces all components to be imported into the single Layout.vue instead of having VuePress manage the components and when they should be loaded.

#### What does the proposed API look like?
By using a specific component name, i.e. AppLayout.vue, SPALayout.vue or Global.vue this component could be loaded first and then the normal Layout.vue or AlternateLayout.vue (or whatever is specified in front matter of the MD file). 

#### How should this be implemented in your opinion?
I am not sure of the internals of VuePress to know the proper approach, but how I handled this was the following:

In Layout.vue:

```
<template>
      <my-header />
      <my-navbar />

      <!-- Using the Layout frontmatter, we will load the appropriate page component -->
      <v-content>
        <component :is="layout" :content="content"></component>
      </v-content>

      <my-footer />
</template>
```

Then, in the script, I have:

```
<script>
import myHeader from "../components/my-header";
import myNavbar from "../components/my-navbar";
import myFooter from "../components/my-footer";

import Default from "./Default";
import Home from "./Home";
import AlternateLayout from "./AlternateLayout";
import Blog from "./Blog";
import Manpage from "./Manpage";

export default {
  components: {
    myHeader,
    myFooter,
    myNavbar,
    Default,
    Home,
    AlternateLayout,
    Blog,
    Manpage
  },
  computed: {
    config() {
      return this.$site.themeConfig;
    },
    layout() {
      return this.$page.frontmatter.layout || "Default";
    },
    content() {
      return this.$page.frontmatter;
    }
  }
</script>
```

This works, but it would be better if VuePress could parse through the files in the /theme folder and generate this rather than having to explicitly include these in the main Layout.vue component.

#### Are you willing to work on this yourself?**
Possibly....though not familiar with internal code.

SGarno:
@bencodezen Tag!   You're it!

meteorlxy:
Related to #878 

One possible solution is to make `LayoutDistributor` customizable, so users / theme authors can use their own `LayoutDistributor`.

cc @ulivz

Laruxo:
[...]

PR #1231, opened 2019-01-26
refactor($core): inject root mixins to root component and make LayoutDistributor functional

meteorlxy:
**Summary**

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes (potentially)
- [ ] No

The `clientRootMixin` should be ssr compatible.

In fact, developers should always be aware of that in Vuepress, but our current `updateMeta.js` of __core__ and `mixin.js` of __docs__ are not so good at that.

So this is not an exact "breaking change".

**Other information:**

Based on #1228.

- Inject root mixins to `<Root>` component, so they are "real" root mixins now.
- Make `LayoutDistributor` functional. It would be lighter-weighted, and the structure of Vuepress App in `vue-devtool` would be much clearer.

Maybe we could have further discussion about this implementation @ulivz


packages/@vuepress/core/lib/app/components/LayoutDistributor.js
```diff
@@ -0,0 +1,19 @@
+import Vue from 'vue'
+import { getLayoutAsyncComponent } from '../util'
+
+Vue.component('Layout', getLayoutAsyncComponent('Layout'))
+Vue.component('NotFound', getLayoutAsyncComponent('NotFound'))
```
meteorlxy: I propose to move these two `Vue.component()` lines to `app.js`.

What do you think about that? @ulivz

meteorlxy:
@ulivz I understand. I would like to express my opinions:

---
`rootMixins` is more reasonable to be injected to the root component, rather than to `LayoutDistributor`, because:

- As its literally meaning
- As how `clientRootMixin` explained in the docs:
  > A path to the mixin file which allow you to control the life cycle of root component.

One possible usecase: user wants to add some properties to the root component via `clientRootMixin`, and accesses them via `this.$root` in some component. It's definitely different to inject them to `LayoutDistributor`.

---
About making `LayoutDistributor` functional:
- `LayoutDistributor` does nothing but conditionally render component. It's a typical functional component.
- Functional component is much cheaper to render
- `LayoutDistributor` should be "transparent" to users, and functional component won't show up in the Vue devtools component tree, so that users would only see what they are using in Vue devtools
- It's a preparation for future features like #1226 . If we could allow users to use custom `LayoutDistributor`, we should make our default `LayoutDistributor` as simple as possible. So we should move those `Vue.component()` codes out of it.

---

Seems like we usually have different opinions. Hope you can accept my opinion, but never mind if you declined. I just want to help make this project better 😉

ulivz:
额，clientRootMixin 不能写在 new Vue 中，路由都没 resolve，ready hook 能拿到啥？先了解好背景再来废话。

Issue #931, opened 2018-10-14
Make / unnecessary at the end of URLs


sergeyfilimonov:
## Feature request

There should be an option to remove / after a URL so that site.com/blog/ would look work as site.com/blog 

#### What problem does this feature solve?

It removes unnecessary and ugly slash after a URL.

#### What does the proposed API look like?

No idea.

#### How should this be implemented in your opinion?

It should be configured in config.js.

#### Are you willing to work on this yourself?**

I would wish to work on it but no as I am not qualified enough for this.

martinsotirov:
Not having a slash at the end would imply this is a file so that won't work in most browsers. You can either have "/blog/" or "/blog.html".

sergeyfilimonov:
> Not having a slash at the end would imply this is a file so that won't work in most browsers. You can either have "/blog/" or "/blog.html".

In most cases this is not the case. Actually I notice no slash in 99,9% of my web browsing experience. For instance, this issue has the URL https://github.com/vuejs/vuepress/issues/931 — no slash at the end.

martinsotirov:
When you're serving index.html files from nested directories, you automatically get a slash at the end. There are ways to configure your server to rewrite those requests but it's definitely not the default behavior. See [this](https://webmasters.stackexchange.com/questions/45267/removing-trailing-slash-for-sub-directory-url-indexing) for an Apache example.

In any case, this is beyond the scope of a static site generator in my opinion.

sergeyfilimonov:
> When you're serving index.html files from nested directories, you automatically get a slash at the end. There are ways to configure your server to rewrite those requests but it's definitely not the default behavior. See [this](https://webmasters.stackexchange.com/questions/45267/removing-trailing-slash-for-sub-directory-url-indexing) for an Apache example.
> 
> In any case, this is beyond the scope of a static site generator in my opinion.

Thanks! Didn't know about this. I guess Nginx works the same. Yes, it is unlikely about a site generator but let's give an opportunity to confirm this to someone else. If after a week the issue will not be replied I'll close it/

ulivz:
Thanks for your interest in VuePress, but it's duplicate of https://github.com/vuejs/vuepress/issues/603.

Issue #720, opened 2018-08-10
failed to redirect from '/routePath' to '/routePath/', maybe the problem is about named routes


Wangszzju:
#### Version
0.14.1 

#### Steps to reproduce
* First，download the git repo of [vuepress](https://github.com/vuejs/vuepress)
* Secondly，run the command
```
npm run dev
```
* visit the URL `http://localhost:8080/config/`，OK，it works
* then, visit the URL `http://localhost:8080/config`

#### What is expected?

The page shoule be redirect to `http://localhost:8080/config/`

#### What is actually happening?

The page does not show anything.

#### My Pull Request Here

[It's my PR here](https://github.com/vuejs/vuepress/pull/722)

#### Other relevant information

I used the source panel and debug step by step.

It seems that when the router got the url `config`，it tries to
```
  // redirect /foo to /foo/
  router.beforeEach((to, from, next) => {
    if (!/(\/|\.html)$/.test(to.path)) {
      next(Object.assign({}, to, {
        path: to.path + '/'
      }))
    } else {
      next()
    }
  })
```
but  the next function had been invoked again and again , and the route.path was still the '/config' rather than '/config/'.

And I uninstall the `vuepress`@0.14.1 and install the version 0.8.4, and the problem does not exist.

I compared them. The main difference came from the statement, in `History.prototype.transitionTo`.
```
var route = this.router.match(location, this.current);
```
when the version of vuepress is 0.14.1, the path prop in the return value of the match method, is `/config`, while it's `/config/` in 0.8.4. But before the statement is executed,  the value of the path prop in the first parameter `location`，is `/config/`.

And another diffrence is that the route object has name prop in the version 0.14.1, while it does not in version 0.8.4

I had not read the source code of `vue-router` and `vuepress`. But I guess there is something wrong with the name prop of the route object. And  I noticed that a route config is created in the dir `temp/routes` when you run the `dev` command.

- Your OS:  macOS Sierra 10.12.6
- Node.js version: 9.8.0
- Browser version:  Chrome  68.0.3440.84
- Is this a global or local install?  local
- Which package manager did you use for the install? yarn

meteorlxy:
It's a known 'bug', or Vuepress is originally designed to have this behavior.

In fact I also think url without trailing slash looks better 👍

ulivz:
Thanks for your interest in VuePress, but it's duplicate of #603, let's move to here to continue discussion.

PR #1060, opened 2018-12-01
✨ Remove '.html' extension from URLs

dcastil:
**Summary**

As discussed in #608 and a few other issues, it would be nice to enhance user experience regarding URLs created by vuepress. I want to make a first step towards this goal with this PR and removed the `.html` extension at the end of all URLs. So the URL `/about.html` gets `/about/`.

However, there is some more work to be done.

1. When I start a project with `vuepress dev` and request a URL without the trailing `/`, the router gets into an infinite loop and a blank website is shown.

2. Generally the trailing `/` should be removed from all URLs. Removing the `.html` extension is only the first step and I wouldn't ship this to production as is.

3. Some refactoring might be needed as a lot of code was built with the URL extension in mind.

I'm open for suggestions and help.

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

If changing the UI of default theme, please provide the **before/after** screenshot:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [X] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [X] Chrome
- [X] Firefox
- [X] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [X] A convincing reason for adding this feature
  → [This comment from awulkan](https://github.com/vuejs/vuepress/issues/608#issue-334865699) has some convincing reasons.
- [ ] Related documents have been updated
- [ ] Related tests have been updated

To avoid wasting your time, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

I'm not sure if I have enough time to go though all the points mentioned in the summary. If anyone wants to pick up from my work, feel free to fork from my repo and continue working on that.

Issue #608, opened 2018-06-22
[Feature request] Cleaner and friendlier URLs


awulkan:
## Feature request
VuePress already has somewhat SEO friendly URLs in the sense that each page has the file name as the URL, but there's still some flaws. Vuepress shows the `.html` extension in the URL and it appends a slash on pages which causes inconsistencies. I suggest we get rid of this for a few reasons listed below.

**TL;DR** (if you don't want to read everything):
I want this:
`mysite.com`
`mysite.com/blog`
`mysite.com/blog#title`
`mysite.com/blog/post`
`mysite.com/blog/post#title`

Instead of this:
`mysite.com`
`mysite.com/blog/`
`mysite.com/blog/#title`
`mysite.com/blog/post.html`
`mysite.com/blog/post.html#title`

#### What problem does this feature solve?

**1.** Migrating to/from VuePress.
The vast majority of all popular static site generators or CMS don't use the extension in the URL nowadays. And if they do, it's not always a .html extension. This causes problems when migrating a website to or from VuePress since search engines like Google will treat the new URL path as a totally new page, causing lost search engine rankings.

**2.** Cleaner, shorter and easier to type.
`mysite.com/blog/my-article` vs `mysite.com/blog/my-article.html`
`mysite.com/blog/my-article#title` vs `mysite.com/blog/my-article.html#title`

**3.** Less inconsitencies.
For example, right now it looks like this for pages and posts:
`mysite.com/blog/#title`
`mysite.com/blog/post.html#title`

This is inconsistent. pages have a slash, posts don't. There is no reason for a page to have  slash since it's practically not a directory of things. The only reason I assume its like that right now is for legacy reasons that are not valid anymore.

**4.** No more ugly workarounds.
Right now I have to put each markdown file in a separate folder and name the file index.md to get similar results. It still doesn't generate a proper URL though since it will add an extra slash at the end even for the pages.

What you get:
`mysite.com/blog/my-article/`
`mysite.com/blog/my-article/#title`

How it should look:
`mysite.com/blog/my-article`
`mysite.com/blog/my-article#title`

#### What does the proposed API look like?
I don't really know yet.

#### How should this be implemented in your opinion?
`mysite.com/blog` <- Directories. (without slash)
As I said previously, each "directory" is actually a page as well, so no slash is needed. It would also make the URLs more consistent across the whole site.
For example, Github doesn't use a slash:
`https://github.com/vuejs`
`https://github.com/vuejs/vuepress`
[...]

PR #1269, opened 2019-02-08
feat($core): redirects for clean urls

ulivz:
## Summary 

In VuePress, we have following convention about routing:

   - `/foo/` means source file is `/foo/{README|index}.md`
   - `/foo.html` means your source file is `/foo.md`

 The original design of VuePress relied on above two styles
 of routing, especially the calculation involved of routes at
// default theme. so we can't easily modify `/foo.html` directly
 to `/foo` (i.e. remove html suffix)

 This  utility handles redirect of clean urls, with this utility, you'll
 get:

 For unknown request `/foo`
   - redirect to `/foo.html` if it exists
   - redirect to `/foo/` if it exists

 For unknown request `/foo/`
   - redirect to `/foo.html` if it exists

If all the above redirect rules don't exist, you'll get a 404

---

**Note that** this PR doesn't provide you the ability to remove the
html suffix at the route level, it just allows you to use clean urls.

cc @octref 

---

Related: #603 
Related: #608
Related: #1060
Close: #720 
Close: #931

### DIFF
--- a/packages/@vuepress/core/lib/app/app.js
+++ b/packages/@vuepress/core/lib/app/app.js
@@ -8,6 +8,7 @@ import appEnhancers from '@internal/app-enhancers'
 import globalUIComponents from '@internal/global-ui'
 import ClientComputedMixin from '@transform/ClientComputedMixin'
 import VuePress from './plugins/VuePress'
+import handleRedirect from './redirect.js'
 
 // built-in components
 import Content from './components/Content.js'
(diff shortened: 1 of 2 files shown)


awulkan:
If I understood this correctly, the routes will still end in .html? Or am I wrong?

ulivz:
I have noted it in the description:

> **Note that** this PR doesn't provide you the ability to remove the
html suffix at the route level, it just allows you to use clean urls.

FYI, we don't plan to support removing the `html` suffix for now(1.x)

awulkan:
> FYI, we don't plan to support removing the `html` suffix for now(1.x)

Oh, so that won't happen until 2.x? Which might not happen for a LONG time. I guess I'll use another static site generator for now then, because I can't afford losing search engine rankings when you eventually do allow non-.html links.

ulivz:
It's up to you.

Shigma:
@awulkan 

Now you can use [plugin-clean-urls](https://v1.vuepress.vuejs.org/plugin/official/plugin-clean-urls.html) if you want your urls to be `/foo/` or even `/foo` instead of `/foo.html`.

nelson6e65:
OMG! Amazing, @Shigma

awulkan:
@Shigma Well done. :)

PR #1308, opened 2019-02-17
Update outbound link icon alignment

gtsiolis:
**Summary**

This will update outbound link icon alignment.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

| Before (header) | After (header) |
| ------ | ------ |
| image | image |

| Before (paragraph) | After (paragraph) |
| ------ | ------ |
| image | image |

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/packages/@vuepress/core/lib/app/components/OutboundLink.vue
+++ b/packages/@vuepress/core/lib/app/components/OutboundLink.vue
@@ -9,4 +9,7 @@
 .icon.outbound
   color #aaa
   display inline-block
+  vertical-align middle
+  position relative
+  top -1px
 </style>


gtsiolis:
@ulivz could you review or forward this?

gtsiolis:
Thanks @ulivz! The goal was *not* to vertically align the icon but actually use `text-bottom` alignment in the first place. I think that both approaches are acceptable but let me know if you think this should vertically align the icon in the middle instead.

Shigma:
> It's also centered when you change the font-size:

I think outbound link will look better if keeping the same size as fonts.

ulivz:
@Shigma Yup but we can do it in another PR, since there are two things.

Shigma:
> @Shigma Yup but we can do it in another PR, since there are two things.

I bet you are right.

gtsiolis:
@ulivz I tried this out with a couple of different font sizes and the suggestion in https://github.com/vuejs/vuepress/pull/1308#pullrequestreview-204816792 looks much better for larger fonts. I've updated the code and the screenshots in the description. Let me know what you think.

gtsiolis:
Thanks @ulivz! 😆

PR #1316, opened 2019-02-19
Code snippet hmr, fix #1309

### LINKED ISSUE #1309: Not rerender when code snippet file changes

#### Version

1.0.0-alpha.39

#### Steps to reproduce

The issue is found by @jiaming743 in #1297. Thanks!

Suppose we have two files, *foo.js* and *foo.md*.

**foo.js**

```js
export default {
    // this is a test file
}
```

**foo.md**

```md
<<< @/foo.js
```

#### What is expected?

When updating *foo.js*, *foo.html* will automatically rerender.

After updating *foo.js*, *foo.html* does not changes.

- Your OS: Windows
- Node.js version: v10.12.0
- Browser version: N/A
- Is this a global or local install? Local
- Which package manager did you use for the install? Yarn

Shigma:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

- support code snippet hmr, fix #1309 
- no more cache on `md.render` but use cache on `md.parse`
- cache on `md.renderer.render` is currently working in progress on [this fork](https://github.com/Shigma/vuepress/tree/code-snippet-hmr) (I think it may be better using another pull request)


Shigma:
### Something to be discussed before merge

We use cache to speed up `md.parse`:

```js
const parse = md.parse
const cache = new LRUCache({ max: 1000 })
md.parse = (src, env) => {
  const key = hash(src + env.relPath)
  const cached = cache.get(key)
  if (cached) {
    return cached
  } else {
    const tokens = parse.call(md, src, env)
    cache.set(key, tokens)
    return tokens
  }
}
```

And these will lead to a problem: **if someone changes one of those `tokens` during render process, it will be a disaster**, because changing one of those `tokens` mean changing cache during render.

Here is an example (markdown/lib/link.js):
```diff
-  return Object.assign({}, token, {
-    tag: 'router-link'
+  return Object.create(token, {
+    tag: { value: 'router-link' }
   })
```

What is the difference? The former use `assign`, which will drop all the prototype methods (like `Token.prototype.attrIndex`, and will lead to an error at next time:

```js
const hrefIndex = token.attrIndex('href')
// Error: token.attrIndex is not a function
```

#### How can we fix it?

By copy all the tokens before `get` and `set`:

```js
md.parse = (src, env) => {
  const key = hash(src + env.relPath)
  const cached = cache.get(key)
  if (cached) {
    return copyTokens(cached)
  } else {
    const tokens = parse.call(md, src, env)
    cache.set(key, copyTokens(tokens))
    return tokens
  }
}

// I actually don't know whether we should
// prevent modification to cached tokens
function copyTokens (tokens) {
  return tokens.map((token) => {
    if (token.children) {
      token.children = copyTokens(token.children)
    }
    return Object.create(token)
  })
}
```

However, I still don't know whether we should do such thing, because I think users should not change those tokens during rendering process.

Shigma:
@BuptStEve Can you check if this pr fix #1309? And what do you think about the question above?

BuptStEve:
> @BuptStEve Can you check if this pr fix #1309? And what do you think about the question above?

It works.
Just some ideas. Maybe in `markdown-loader`, we call `md.parse` to check the cache and we cache the result of `md.render`?

Shigma:
> Just some ideas. Maybe in `markdown-loader`, we call `md.parse` to check the cache and we cache the result of `md.render`?

It will not work. Because `addDep` was called in the rendering process.

Shigma:
It may be better splitted into some smaller PRs. Closed.

PR #1322, opened 2019-02-20
handle error on loading module, fix #1296

### LINKED ISSUE #1296: Error in local theme index.js should not be ignored and reported as "cannot resolve plugin ..."

#### Version

1.0.0-alpha.36 

#### Steps to reproduce

```
.vuepress
├─ config.js
├─ theme
   ├─ index.js
   ├─ and other files ...
```

Assuming that an error was encounted in `index.js`, vuepress will leave a message `cannot resolve plugin .../.vuepress/theme`. But this behavior is rather strange and may mislead user with an indication that the theme is **not found**. Also, I think errors in local theme index file should not be simply ignored.

#### What is expected?

- stop the process like an error was encounted in the `config.js`, or
- provide a better warning like `an error was encounted in ......`

`cannot resolve plugin .../.vuepress/theme`

- Your OS: N/A
- Node.js version: N/A
- Browser version: N/A
- Is this a global or local install? N/A
- Which package manager did you use for the install? N/A

Shigma:
**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

---

### Old non-string request resolution

Try to resolve the request as absolute/relative/dependency path and import them in chain. Use the first result as the resolved module. If any error was encounted, respective resolution will be skipped.

### What are the drawbacks?

1. It's not consistent with the implementation of Node.js. Node.js will take the first existing file as the final entry no matter it has errors or not.
2. Referenced in #1296. A plain error log `cannot resolve plugin` may mislead user with an indication that the theme or plugin is **not found** instead of **wrong**.
3. Whether the error is in the theme or plugin, the same error with `plugin` is displayed.

### This PR's implementation

Try to resolve the request as absolute/relative/dependency path in chain.

- If an entry file was found, vuepress will some the try-chain and try to load the file.
  - If the entry file is successfully loaded, apply it as a plugin/theme.
  - If an error was encounted, vuepress will leave a error log `en error was encounted in plugin "xxx"`.
- If all possible files do not exist, vuepress will leave a error log `cannot resolve plugin "xxx"`.
- If similar problem was encounted in theme entry file resolution, the error log will replace `plugin` with `theme`.

### Breaking change

The probability that a breaking change occurs is rather small. Supposing there are two "entry" files with the same name. One is in the relative path (call it *foo1*), the other is a dependency of the project (call it *foo2*). Supposing *foo1* has an error in it, and *foo2* is normal:

```js
// foo.js (foo1)
throw new Error()
```
```js
// foo.js (foo2)
module.exports = {
  name: 'foo'
}
```

*foo2* will be loaded with previous methods, while no module will be loaded in this PR.

---

N/A


Shigma:
Moved to #1413.

PR #1403, opened 2019-03-05
feat($core): refine clientDynamicModules

### LINKED ISSUE #1338: Only arrow functions work properly as arguments to Pagination Plugin (normal functions eval to undefined)

#### Version
`vuepress/1.0.0-alpha.39`

#### Steps to reproduce

The shortest reproduction is to run these two commands in your developer console:

```js
eval('function f(a) { return a }')
// undefined
eval('(a) => a')
// (a) => a
```

To see this in the context of vuepress, create a blank project that uses `@vuepress/pagination`, and then pass a *normal* function (not an arrow function) to either `postsSorter` or `postsFilter`. Since that plugin uses `eval` to recover those functions:
https://github.com/vuejs/vuepress/blob/master/packages/%40vuepress/plugin-pagination/enhanceAppFile.js#L8

`undefined` will be returned, and the build will fail, saying that undefined isn't a function.

Similarly, if someone defines their `postsSorter` or `postsFilter` as an *anonymous* normal function, `eval` will throw a SyntaxError:

```js
eval('function () {}')
// Uncaught SyntaxError: Unexpected token ( at <anonymous>:1:1
```

#### What is expected?
The plugin should successfully recover either kind of function.

The plugin throws an error, saying that undefined is not a function.

- Your OS: `Ubuntu 18.04` 
- Node.js version: `node-v11.8.0`
- Browser version: `Chrome 72 on Linux`
- Is this a global or local install? `global`
- Which package manager did you use for the install? `npm`

Shigma:
**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### A new property for every client dynamic modules: data

If specified, it will be stringified as `export default` of the dynamic module.

### A simple example (will fix #1338)

```js
// index.js
module.exports = (options) => {
  clientDynamicModules: {
    name: 'pagination.js',
    data: {
      postsFilter, // it can be a function
      postsSorter, // it can be a function
    },
  },
}
```
```js
// enhanceApp.js
import pagination from '@dynamic/pagination'
const { postsFilter, postsSorter } = pagination
```

### Related documents and tests have been updated

- add a test on `stringify.ts`
- refine documents on `clientDynamicModules`

**Other information:**


Shigma:
Maybe it sounds better with `default` instead of `data`. I just don't know.

Shigma:
Maybe is not a good idea to put `stringify-object`. So I have wrote a package for that.

[@shigma/stringify-object](https://www.npmjs.com/package/@shigma/stringify-object)

Issue #1503, opened 2019-04-01
Site config base will always get / (bug in v1.0.0-alpha.46)


afeiship:
// routerBase will always get '/'
- version: v1.0.0-alpha.44 is `correct`
- version: v1.0.0-alpha.46 is `wrong`

```js
// @vuepress/core/lib/client/app.js
export function createApp (isServer) {
  const routerBase = typeof window !== 'undefined'
    ? window.__VUEPRESS_ROUTER_BASE__
    : (siteData.routerBase || siteData.base)

  const router = new Router({
    base: routerBase,
```

timaschew:
With version v1.0.0-alpha.44 I was using `base` in my config which has worked.
Now in version v1.0.0-alpha.46 it doesn't work anymore, I get a 404 if I want to access the url which is specified via `base`. But the page works if I manually open `/`.

timaschew:
For anyone else: to fix this you need to downgrade not vuepress itself, but the core:

`npm i @vuepress/core@1.0.0-alpha.44`

afeiship:
@timaschew thanks!

Shigma:
@ulivz 

Can you look into this regression?

ulivz:
Why did you always get `/`?

PR #1504, opened 2019-04-01
fix: site base get /

afeiship:
See: https://github.com/vuejs/vuepress/issues/1503


Shigma:
Thanks for the PR. However, it is not a good approach for me.

```js
const routerBase = typeof window !== 'undefined'
  ? window.__VUEPRESS_ROUTER_BASE__
  : undefined

const router = new Router({
  base: routerBase || siteData.routerBase || siteData.base
```

PR #1509, opened 2019-04-02
Fix #1505 - Snippets should not split path on space

### LINKED ISSUE #1505: Snippet file "not found" error if the path has a space in it

## Bug report
Including a snippet fails if there is a space anywhere in the path name. This snippet:

```
<<< @/docs/dev/tools/Test File.js
```

Results in an error showing the cut-off filename:

```
Not found: /Users/myuser/dev/my-cool-website/docs/dev/tools/Test
```

This happens if the space is in the `@` alias (for cwd) -- I can rename the files under `docs/`, but when this builds in my CI env, there's a space that is outside of my control, which is really unfortunate!

#### Version
I am using 0.14.10.

Looking at the snippet code in the 1.x branch, I believe 1.x would be affected as well.

#### Steps to reproduce
1. Create a page & [include the snippet](https://vuepress.vuejs.org/guide/markdown.html#import-code-snippets) `Test Doc.js`
1. Create a `Test Doc.js` file
1. View the page & see the error.

#### What is expected?
The snippet should be included in the page.

The snippet is not being included in the page.

- Your OS: OS X 10.13.6
- Node.js version: v11.12.0
- Browser version: n/a
- Is this a global or local install? Local 
- Which package manager did you use for the install? Yarn
- Does this issue occur when all plugins are disabled? Yes

nie7321:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No -- In any case, I don't *think* it does, but see [my earlier question](https://github.com/vuejs/vuepress/issues/1505).

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:** I hope the test is wanted. I'm (selfishly) submitting this to the 0.x branch, since that's the version I use, but I think this would need to be applied to 1.x too.

### DIFF
--- a/lib/markdown/snippet.js
+++ b/lib/markdown/snippet.js
@@ -24,7 +24,7 @@ module.exports = function snippet (md, options = {}) {
     const start = pos + 3
     const end = state.skipSpacesBack(max, pos)
     const rawPath = state.src.slice(start, end).trim().replace(/^@/, root)
-    const filename = rawPath.split(/[{\s]/).shift()
+    const filename = rawPath.split(/{/).shift()
     const content = fs.existsSync(filename) ? fs.readFileSync(filename).toString() : 'Not found: ' + filename
     const meta = rawPath.replace(filename, '')
 
--- a/lib/markdown/snippet.js
+++ b/lib/markdown/snippet.js
@@ -24,7 +24,7 @@ module.exports = function snippet (md, options = {}) {
     const start = pos + 3
     const end = state.skipSpacesBack(max, pos)
     const rawPath = state.src.slice(start, end).trim().replace(/^@/, root)
-    const filename = rawPath.split(/{/).shift()
+    const filename = rawPath.split(/{/).shift().trim()
     const content = fs.existsSync(filename) ? fs.readFileSync(filename).toString() : 'Not found: ' + filename
     const meta = rawPath.replace(filename, '')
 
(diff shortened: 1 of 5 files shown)


lib/markdown/snippet.js
```diff
@@ -24,7 +24,7 @@ module.exports = function snippet (md, options = {}) {
     const start = pos + 3
     const end = state.skipSpacesBack(max, pos)
     const rawPath = state.src.slice(start, end).trim().replace(/^@/, root)
-    const filename = rawPath.split(/[{\s]/).shift()
+    const filename = rawPath.split(/{/).shift()
```
Shigma: In that case, we need to add a `.trim()` because `rawPath` may be something in the form of `@/path/to/file {1}`.

nie7321: Ah, that's why it was also allowing a split on space -- sorry, I should have realized that myself!

I've updated the jest test I added so it checks the highlighting works when there's a space between the path & highlight as well.

ulivz:
Thanks, and BTW, next time please follow the [conventional commits guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.3/).

ulivz:
Also, could you help update the 1.x branch?

nie7321:
> Also, could you help update the 1.x branch?

Certainly! I'll get a PR in soonish.

PR #1618, opened 2019-05-23
feat: add async load of config

### LINKED ISSUE #1185: Support asynchronous config for generated sidebar
Using a generated sidebar that is based on dynamic content. I'm fetching content before running vuepress build and as a workaround I now write the data needed for the sidebar in a JSON file and then synchronously read it. Instead it would be better to fetch the content and generate the sidebar in one go so no temporary file is needed.

Two options, either the config export could return a Promise or the sidebar value can. E.g:
```js
const fetchContent = () => Promise.resolve();
const generateSidebar = () => Promise.resolve({});

module.exports = {
  sidebar: fetchContent.then(generateSidebar),
}
```
Going with the sidebar example vuepress could check if the `sidebar` value is a Promise or not before building, which would also be backwards compatible.
Maybe, depending on the complexity of the change since I have no idea how far changing the code goes to make a part of the config resolution async.

romulocintra:
**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

fix #1185

Needs more convenient tests

### DIFF
--- a/packages/@vuepress/core/lib/node/loadConfig.js
+++ b/packages/@vuepress/core/lib/node/loadConfig.js
@@ -28,9 +28,12 @@ module.exports = function loadConfig (vuepressDir, bustCache = true) {
   } else if (fs.existsSync(configTomlPath)) {
     siteConfig = parseConfig(configTomlPath)
   } else if (fs.existsSync(configPath)) {
-    siteConfig = require(configPath)
+    if (typeof require(configPath).then === 'function') {
+      siteConfig = require(configPath).then((config) => config)
+    } else {
+      siteConfig = require(configPath)
+    }
   }
-
   return siteConfig
 }
 
--- a/packages/@vuepress/core/lib/node/loadConfig.js
+++ b/packages/@vuepress/core/lib/node/loadConfig.js
@@ -4,7 +4,7 @@
  * Module dependencies.
  */
 
-const { fs, path } = require('@vuepress/shared-utils')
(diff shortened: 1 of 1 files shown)


f3ltron:
Thank's for your work @romulocintra ! I know it's been a while since you create the pr.

I will have to check a bit about async stuff here. It's look good anyway.

packages/@vuepress/core/lib/node/loadConfig.js
```diff
@@ -28,9 +28,12 @@ module.exports = function loadConfig (vuepressDir, bustCache = true) {
   } else if (fs.existsSync(configTomlPath)) {
     siteConfig = parseConfig(configTomlPath)
   } else if (fs.existsSync(configPath)) {
-    siteConfig = require(configPath)
+    if (typeof require(configPath).then === 'function') {
+      siteConfig = require(configPath).then((config) => config)
```
f3ltron: if you can add a catch with a pretty error message it will be good for me.

use logs in shared utils package to log the error

romulocintra: I will

romulocintra: Hi @f3ltron  not to much time to review and add a proper commit but updated with catch and logger.error message

f3ltron:
tank's for your work ! 👍 i validate it

PR #1640, opened 2019-06-04
fix($last-updated): use file author time instead of submodule commit time (1.x)

DvdGiessen:
**Summary**
When using a Git submodule, the "last updated" time for all files within that submodule is incorrect.

This happens because the `git log` command, when operating from a working directory that is part of the parent repository, only operates on that parent repository. As a result, any query for the log for a specific file that is located in a submodule will return the log for the submodule entry itself.

So, for example, if we have a submodule `/foo` and a file `/foo/bar.md`, then even if `bar.md` was last updated many years ago, if we commited a update of the submodule five minutes ago that will cause Vuepress to show that `bar.md` was last updated 5 minutes ago, while in fact it hasn't.

This pull request resolves this bug by first changing the current working directory to the directory containing the file, so that if the file is part of a submodule, `git log` accesses the log of the submodule and retrieves the correct last updated time from that files history instead of only considering the submodule history of the parent repository. This of course also works the same for repositories without submodules so does not require any additional checks or workarounds.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE
- [x] Checked generated code in text editor

**Other information:**
I've also opened a pull request with the exact same changes for `0.x`, see #1639.

### DIFF
--- a/packages/@vuepress/plugin-last-updated/index.js
+++ b/packages/@vuepress/plugin-last-updated/index.js
@@ -1,3 +1,4 @@
+const path = require('path')
 const spawn = require('cross-spawn')
 
 module.exports = (options = {}, context) => ({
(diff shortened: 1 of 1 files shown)


f3ltron:
hello @DvdGiessen thank's for the work and time ! Can you maybe provide a test for it ?

DvdGiessen:
I'm afraid I won't be able to provide a test for this change any time soon; there are no existing tests for this functionality for me to quickly update and unfortunately I currently don't have much time to spend on writing a complete test for this plugin. Sorry!

I agree that is would be desirable to have tests for this functionality, so anyone else wants to take a look I welcome them very much to pick my small contribution and improve upon it with some proper tests. :)

PR #1731, opened 2019-07-31
 fix(Util): use any custom protocol for outboundRE

xr0master:
Little improvement for vuejs#1677

I think we can use any custom protocol for external link.


ekoeryanto:
I think the regex missed slashes

PR #1745, opened 2019-08-06
feat: documentation linting

sarahdayan:
**Summary**

This PR adds content linting to the documentation of VuePress, using [textlint](https://github.com/textlint/textlint).

Some rules (the more subjective ones) are set with a warning severity, meaning they wouldn't make a CI fail. It also leaves the possibility of progressively fixing things.

The PR is quite large because it also embarks a first run at fixing issues that were revealed by textlint.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Content linting was added as a pre-commit hook. I changed the way the other pre-commit hook was wired so the whole thing works properly across the monorepo.

❓**Question:** for now, the linting precommit hooks are split into two:

- a local one in `docs/` (the one created in this PR)
- a global one at the root (to lint JavaScript and Vue files)

They're triggered by running the global precommit first, then the local ones with `lerna run`.
Ideally, it's better to have a command per repo. Do we think it's worth it?


sobolevn:
@sarahdayan how about Chinese and Russian languages? Are they supported?

sarahdayan:
> @sarahdayan how about Chinese and Russian languages? Are they supported?

Unfortunately they're not. There are some rules for Japanese out there, but if we want to support Chinese and Russian we'll have to create new rules.

PR #1759, opened 2019-08-08
Improve CI process

### LINKED ISSUE #1748: Improve CircleCI workflow
## Feature request

I've noticed that some improvements could be added to the CI process:

- Split the single main process into multiple jobs.
   **Why ?**
   - From a github PR, it would be easier to understand what's wrong when CI fails.
We'd have this:
Sans titre

      Instead of this:
Capture d’écran 2019-08-06 à 11 06 43

   - Ci process would be faster because we could parallelize some jobs.

- Saving cache should rely on [`yarn.lock` checksum instead of `package.json`](https://circleci.com/docs/2.0/yarn/#caching).

- We should use `--frozen-lockfile` flag when installing deps with yarn (on CI).

- Current docker image is `circleci/node:10.15.2` but I think we should use `circleci/node:latest` as I don't see any compatibility note telling that we shouldn't use latest versions of node 🙂

- Current `lint` command fix errors (so it never fails). The CI process should never change source code. CI lint command has to be replaced by a `lint:check` that just check if code is correctly formatted and exit with error code if necessary. 

Yes 👍I'm on vacation for the moment, but I'll do it as soon as I get back 🏖

kefranabg:
**Summary**

Fix #1748

New CI workflow introduced in this PR:

Capture d’écran 2019-08-08 à 12 13 19

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [x] Other, please describe:
Improving CI process

- [ ] Yes
- [x] No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,36 +2,122 @@
 #
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 #
-version: 2
+version: 2.1
+
+defaults: &defaults
+  working_directory: ~/project
+  docker:
+    - image: circleci/node:latest
+
 jobs:
-  build:
-    docker:
-    # specify the version you desire here
-    - image: circleci/node:10.15.2
 
-    # Specify service dependencies here if necessary
-    # CircleCI maintains a library of pre-built images
-    # documented at https://circleci.com/docs/2.0/circleci-images/
-    # - image: circleci/mongo:3.4.4
+  #------------------------------------------------------------
+  # 1. Install dependencies
+  #------------------------------------------------------------
 
-    working_directory: ~/repo
-
+  install-dependencies:
+    <<: *defaults
     steps:
-    - checkout
+      - checkout
 
-    # Download and cache dependencies
-    - restore_cache:
-        keys:
-        - v1-dependencies-{{ checksum "package.json" }}
-        # fallback to using the latest cache if no exact match is found
-        - v1-dependencies-
+      - restore_cache:
+          keys:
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,6 +21,7 @@
     "show-help": "yarn workspace docs show-help",
     "register-vuepress": "lerna exec --scope vuepress -- yarn link",
     "lint": "eslint packages --fix --ext .js,.vue",
```
sarahdayan:
```suggestion
     "lint": "yarn lint:check --fix",
```

.circleci/config.yml
```diff
@@ -2,36 +2,122 @@
...
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: 'Run tests'
+          command: yarn test
 
+  run-linter-check:
```
sarahdayan: We could also add the text linting (remark and textlint). Most current issues are warnings so they won't break the build anyway.

kefranabg: 👍  added in [#919d490](https://github.com/vuejs/vuepress/pull/1759/commits/919d490f4955078640aee7f019c9e9c59eb180f9)

f3ltron:
@kefranabg can you resolve the conflict and @sarahdayan valid the changes :). That i can add it to the release maybe ?

f3ltron:
you will need to resolve conflict here @kefranabg

PR #1778, opened 2019-08-15
Add $page.lastUpdatedTstamp

psi-4ward:
**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [x] Edge
- [ ] IE

**Other information:**

The `$page.lastUpdated` generated from _plugin-last-updated_ helds the Date in the localized format. Other plugins like sitemap can hardly work with this value in a mathematics manner so it's necessary to also expose the Timestamp.

See https://github.com/ekoeryanto/vuepress-plugin-sitemap/issues/11

### DIFF
--- a/packages/@vuepress/plugin-last-updated/index.js
+++ b/packages/@vuepress/plugin-last-updated/index.js
@@ -10,6 +10,7 @@ module.exports = (options = {}, context) => ({
         ? transformer(timestamp, $lang)
         : defaultTransformer(timestamp, $lang)
       $page.lastUpdated = lastUpdated
+      $page.lastUpdatedTstamp = timestamp
     }
   }
 })
--- a/packages/@vuepress/plugin-last-updated/index.js
+++ b/packages/@vuepress/plugin-last-updated/index.js
@@ -10,7 +10,7 @@ module.exports = (options = {}, context) => ({
         ? transformer(timestamp, $lang)
         : defaultTransformer(timestamp, $lang)
       $page.lastUpdated = lastUpdated
-      $page.lastUpdatedTstamp = timestamp
+      $page.lastUpdatedTimestamp = timestamp
     }
   }
 })


f3ltron:
Hello @psi-4ward  ! Thank's for you PR

packages/@vuepress/plugin-last-updated/index.js
```diff
@@ -10,6 +10,7 @@ module.exports = (options = {}, context) => ({
         ? transformer(timestamp, $lang)
         : defaultTransformer(timestamp, $lang)
       $page.lastUpdated = lastUpdated
+      $page.lastUpdatedTstamp = timestamp
```
f3ltron: I think we can add the full name here lastUpdatedTimestamp lastUpdatedTimeStamp for me it could a lot more clearer

psi-4ward: Imho is tstamp a very common term for the Unix (like) time stamp but I will adjust it

f3ltron: Maybe others can answer too. Actually i dont see any short write of variable inside vuepress.

psi-4ward:
Updated to `lastUpdatedTimestamp`

psi-4ward:
The transformer gets used to format`page.lastUpdated` but probably we need other date-formats too - see the [vuepress-sitemap-plugin](https://github.com/ekoeryanto/vuepress-plugin-sitemap/blob/master/index.js#L54). A correct sitemap.xml needs an ISO-String which can only hardly derived from the local date string.

kefranabg:
What you mean is that you need to have two types of date format in the same page right?

kefranabg:
What you want seems very specific to your need. Could [theme inheritance](https://v1.vuepress.vuejs.org/theme/inheritance.html#motivation) solve your problem?

psi-4ward:
> What you mean is that you need to have two types of date format in the same page right?

Not necessarily in the _same_ page. More generic there sould be a possibility to have the date in a machine-readable format. The sitemap-plugins generates a XML of all pages with ISO-Date of the laste-updated date. Another usecase would be a smth like an article-list (blog) with dates like "1 day ago".

psi-4ward:
https://github.com/ekoeryanto/vuepress-plugin-sitemap/issues/11#issuecomment-544226602

psi-4ward:
Issue still valid: https://github.com/ekoeryanto/vuepress-plugin-sitemap/issues/11

psi-4ward:
> Could you help update the relevant documentation?

Can you give me a hint where to document this?

ulivz (approved):
LGTM

Issue #1818, opened 2019-08-30
Accessibility: Navbar dropdown can't be used with just a keyboard


missmatsuko:
- [x] I confirm that this is an issue rather than a question.

## Bug report
There doesn't seem to be any way to access dropdown items in the navbar using keyboard navigation.

#### Steps to reproduce
1. Go to <https://vuepress.vuejs.org/>
2. Tab through the navbar menu items
3. The navbar menu item with a dropdown ("languages") is completely skipped. There's no way to select a language from the navbar using keyboard navigation.

#### What is expected?
Someone using a keyboard to navigate the site should be able to access dropdown items (and in this case, change the site's language).

#### What is actually happening?
Someone using a keyboard to navigate the site can't access dropdown items.

#### Other relevant information

- Output of `npx vuepress info` in my VuePress project:

edisdev:
Hi Sir  @f3ltron ,  I opened pull request for this issue. pr link : https://github.com/vuejs/vuepress/pull/1837

f3ltron:
Hello @edisdev i will check for it top !

PR #1837, opened 2019-09-05
fix($theme-default): Make navbar dropdown links accessible

edisdev:
**Summary**

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

Hi !
I've edited the accessibility of items on the navbar. (for  this issue https://github.com/vuejs/vuepress/issues/1818 )

### DIFF
--- a/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
+++ b/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`DropdownLink renders dropdown link. 1`] = `
-<div class="dropdown-wrapper"><button type="button" aria-label="Languages" class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></button>
+<div class="dropdown-wrapper"><button type="button" aria-label="Learn More Select" class="dropdown-title"><span class="title">Learn More</span> <span class="arrow right"></span></button>
   <ul class="nav-dropdown" style="display: none;" name="dropdown">
     <li class="dropdown-item">
       <!----> <a class="nav-link">Guide</a></li>
--- a/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
+++ b/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`DropdownLink renders dropdown link. 1`] = `
-<div class="dropdown-wrapper"><a class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></a>
+<div class="dropdown-wrapper"><a aria-label="Languages" class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></a>
   <ul class="nav-dropdown" style="display: none;" name="dropdown">
     <li class="dropdown-item">
       <!----> <a class="nav-link">Guide</a></li>
--- a/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
+++ b/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
@@ -1,12 +1,12 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`DropdownLink renders dropdown link. 1`] = `
-<div class="dropdown-wrapper"><a aria-label="Languages" class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></a>
+<div class="dropdown-wrapper"><a aria-label="Languages" tabindex="8" class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></a>
   <ul class="nav-dropdown" style="display: none;" name="dropdown">
     <li class="dropdown-item">
-      <!----> <a class="nav-link">Guide</a></li>
+      <!----> <a class="nav-link" tabindex="8">Guide</a></li>
     <li class="dropdown-item">
-      <!----> <a class="nav-link">Config Reference</a></li>
+      <!----> <a class="nav-link" tabindex="8">Config Reference</a></li>
   </ul>
 </div>
 `;
--- a/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
+++ b/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
@@ -6,7 +6,7 @@ exports[`DropdownLink renders dropdown link. 1`] = `
     <li class="dropdown-item">
       <!----> <a class="nav-link">Guide</a></li>
     <li class="dropdown-item">
-      <!----> <a tabindex="-1" class="nav-link">Config Reference</a></li>
+      <!----> <a class="nav-link">Config Reference</a></li>
   </ul>
 </div>
 `;
--- a/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
+++ b/packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
@@ -1,12 +1,12 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`DropdownLink renders dropdown link. 1`] = `
-<div class="dropdown-wrapper"><a aria-label="Languages" tabindex="8" class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></a>
+<div class="dropdown-wrapper"><button type="button" aria-label="Languages" class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></button>
   <ul class="nav-dropdown" style="display: none;" name="dropdown">
     <li class="dropdown-item">
-      <!----> <a class="nav-link" tabindex="8">Guide</a></li>
+      <!----> <a class="nav-link">Guide</a></li>
     <li class="dropdown-item">
-      <!----> <a class="nav-link" tabindex="8">Config Reference</a></li>
+      <!----> <a tabindex="-1" class="nav-link">Config Reference</a></li>
   </ul>
 </div>
 `;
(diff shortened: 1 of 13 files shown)


packages/@vuepress/theme-default/__tests__/components/__snapshots__/DropdownLink.spec.js.snap
```diff
@@ -1,12 +1,12 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`DropdownLink renders dropdown link. 1`] = `
-<div class="dropdown-wrapper"><a class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></a>
+<div class="dropdown-wrapper"><a aria-label="Languages" tabindex="8" class="dropdown-title"><span class="title">VuePress</span> <span class="arrow right"></span></a>
   <ul class="nav-dropdown" style="display: none;" name="dropdown">
     <li class="dropdown-item">
-      <!----> <a class="nav-link">Guide</a></li>
+      <!----> <a class="nav-link" tabindex="8">Guide</a></li>
```
missmatsuko: We should try to avoid the use of positive values for tabindex if possible:
https://webaim.org/techniques/keyboard/tabindex

packages/@vuepress/theme-default/components/DropdownLink.vue
```diff
@@ -5,7 +5,9 @@
   >
     <a
```
missmatsuko: The NavLink element should be <a> only when it acts as a link (e.g. has a valid `href` value). If it's acting as a trigger to open a dropdown navigation, it should be a <button>. This should probably remove the need to set an explicit tabindex (<a> without href is not focusable, but <button> is focusable, at least on Chrome).

packages/@vuepress/theme-default/components/DropdownLink.vue
```diff
@@ -149,7 +160,8 @@ export default {
 @media (min-width: $MQMobile)
   .dropdown-wrapper
     height 1.8rem
-    &:hover .nav-dropdown
+    &:hover .nav-dropdown,
+    &:focus-within .nav-dropdown
```
missmatsuko: Opening the submenu on focus-within works ok, but it could be annoying if there's a navbar with many long submenus (i.e. you'd have to tab through every single item to get past the navbar). Adding a [skip navigation link](https://webaim.org/techniques/skipnav/) could help with that (and probably should also be added regardless), but a keyboard user would still have to tab through every submenu item on the way to the one they actually want.

The [Adobe Mega Menu](https://adobe-accessibility.github.io/Accessible-Mega-Menu/) is a good example of the expected keyboard interactions and aria markup:

f3ltron: thx for your explanation @missmatsuko

edisdev: @missmatsuko Thank you for comment and review :) You are right, Maybe it can make to focus submenus when key press the enter. Or, can be added focusable property to config file. What do you think about this ?

missmatsuko: I think there doesn't need to be an tabindex attribute at all. Can the link element be changed to a <button> if it has children submenu items, and an anchor if it has an href?

edisdev: I made testing when ı remove tabindex, element can not focusable in safari and firefox.

edisdev: I made testing when I remove tabindex, element can not focusable in safari and firefox. @missmatsuko

missmatsuko:
Is it because the element is an <a> with no href attribute set? The element
should be focusable if this is changed to a <button> element. A button is
more appropriate here anyways because it isn't a link to a page - it should
be used to open/close the submenu.

edisdev:
Unluckily, <a> element has href property but it can not focusable. But as you say , can change dropdown open/close element with button. @missmatsuko

Screen Shot 2019-09-09 at 16 24 33

Screen Shot 2019-09-09 at 16 22 44

edisdev:
What do you think about this ? @f3ltron

missmatsuko:
The anchor with javascript in the href should be a <button> element because it doesn't act as a link, it opens a submenu.
[Link Behavior](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/links/link-behavior/)

Also sorry for breaking out of the thread, I didn't know replying by email would do that. 😬

edisdev:
No problem :) @missmatsuko As i said, I will turn the anchor a button element. I think, focusable property can be optional. As you said, if submenus has so many element, It may not be comfortable and useful.

edisdev:
Hi! @f3ltron @kefranabg . I did an new improvement, according last those described. I think pull request is ready of completely.

f3ltron:
@missmatsuko are you ok with changes ? As i am not so good with accessibility. I will try but i have no real idea if it's good or not

missmatsuko:
I'm confused why `notFocusable` is a config setting. Nav items should just not be focusable if it's hidden (i.e. in a closed submenu).

edisdev:
I thought if user wants to a item not focusable, can be need to it. But, i can remove this setting if you want to. All menu items are become focusable. @missmatsuko

missmatsuko:
Oh I see, I think it's a setting so the **developer** can limit some items to not be focusable because of what I said about how it could be a lot of tabbing for the **user** if there's multiple submenus in the nav. Is that right?

Actually, all items that are accessible for a mouse user should be available for a keyboard user. That's the root of the issue, that keyboard users can't use all the navigation items that a mouse user can.

Ideally, the submenus should open for keyboard users when they hit `enter` or `down` while the parent item is focused, and not on `tab` (that's how `focus-within` works though). Tab should go through the parent items when submenus are closed, or submenu items when the submenu is open. It's how the example I previously linked works.

edisdev:
@missmatsuko  Thank you so much for your help  🤗😊🙏

edisdev:
> Looks good!
Pr is approved :) @kefranabg

kefranabg:
I'll do a review ASAP. Thanks for your work 💪

packages/@vuepress/theme-default/components/DropdownLink.vue
```diff
@@ -33,14 +35,17 @@
...
-              <NavLink :item="childSubItem"/>
+              <NavLink
+                @keypressTab="focusoutDropdown((index + 2) * (subIndex + 1))"
+                :item="childSubItem"/>
             </li>
           </ul>
 
           <NavLink
             v-else
+            @keypressTab="focusoutDropdown(index + 1)"
```
kefranabg: I don't think that `focusoutDropdown` is a correct name for what the function is doing, right?

packages/@vuepress/theme-default/components/DropdownLink.vue
```diff
@@ -33,14 +35,17 @@
             <li
               class="dropdown-subitem"
               :key="childSubItem.link"
-              v-for="childSubItem in subItem.items"
+              v-for="(childSubItem, subIndex) in subItem.items"
             >
-              <NavLink :item="childSubItem"/>
+              <NavLink
+                @keypressTab="focusoutDropdown((index + 2) * (subIndex + 1))"
```
kefranabg: I don't understand the `+ 2`? Could you explain? Why not just `+ 1`?

kefranabg:
I also found a bug:

The dropdown will immediately close when the last child element of it is focused

edisdev:
Hi @kefranabg :) Thank you so much review, i resolved your comments :) And also i fixed your said bug. 👍

packages/@vuepress/theme-default/components/DropdownLink.vue
edisdev: I'm sorry, There was a bug here. I resolved this. 👍

packages/@vuepress/theme-default/components/DropdownLink.vue
```diff
@@ -68,9 +73,38 @@ export default {
...
+      return this.item.ariaLabel || this.item.text
+    }
+  },
+
   methods: {
     toggle () {
       this.open = !this.open
+    },
+
+    getSubItemIndex (index, subIndex) {
```
kefranabg: This part is kind of hard to read/understand, I'll try to find something easier for this part

kefranabg:
We should also add a example for `aria-label` in [this documentation page](https://vuepress.vuejs.org/theme/default-theme-config.html#navbar-links)

packages/@vuepress/theme-default/components/DropdownLink.vue
edisdev: I too will try to find to a more easy solution.

edisdev:
@kefranabg i will resolve immediately your comments. 👍

edisdev:
@kefranabg  I added a example for `aria-label` in [documentation page] ( https://vuepress.vuejs.org/theme/default-theme-config.html#navbar-links ) . But, i could not write description for example for chinese, because i don't know chinese. :(

edisdev:
@kefranabg  I added a example for `aria-label` in [documentation page]( https://vuepress.vuejs.org/theme/default-theme-config.html#navbar-links) . But, i could not write description for Chinese example, because i don't know Chinese. :(

packages/docs/docs/theme/default-theme-config.md
```diff
@@ -72,6 +72,25 @@ module.exports = {
 }
 ```
 
+You can change `aria-label` of dropdown button to with a custom aria-label. Default value is text property. if the dropdown menu consists of language items., default value is 'Select language'
```
kefranabg: I don't think we need an entire section for `aria-label`;) What I meant was just adding `ariaLabel: 'Language Menu'` in existing examples in this page. And I think you can do the same for Chineese

kefranabg:
@edisdev You'll find my proposition in the last commit  to simplify the close dropdown algorithm. What do you think?

edisdev:
I think code is more readable  and clear now :) Thank you :)
> @edisdev You'll find my proposition in the last commit to simplify the close dropdown algorithm. What do you think?

packages/docs/docs/theme/default-theme-config.md
edisdev: Okey, i wil add that in existing examples :)

edisdev:
Thank so much you too :) @kefranabg

PR #1900, opened 2019-09-30
Code block to dynamically register a component

garychou:
As someone relatively new to JS + Vue, the existing sample code for dynamic imports didn't tell me how to actually register the component dynamically, which is the likely case for importing a module.

The proposed code block does both: it dynamically imports the module and also registers the component.

**Summary**

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

**Other information:**

### DIFF
--- a/packages/docs/docs/guide/using-vue.md
+++ b/packages/docs/docs/guide/using-vue.md
@@ -26,6 +26,18 @@ export default {
 </script>
 ```
 
+or
+
+``` vue
+<script>
+export default {
+  components: {
+    'module': () => import('name of module')
+  }
+}
+</script>
+```
+
 ## Templating
 
 ### Interpolation
--- a/packages/docs/docs/guide/using-vue.md
+++ b/packages/docs/docs/guide/using-vue.md
@@ -30,7 +30,7 @@ If your module `export default` a Vue component, you can register it dynamically
 
 ```vue
 <template>
-  <component v-if="currentTabComponent" :is="currentTabComponent"></component>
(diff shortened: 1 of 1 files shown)


kefranabg:
Hi @garychou,

Thanks for your work.

I don't understand why this example would be necessary? The main purpose of this part is to show that dynamic import fix errors for external libs trying to access window.

Your example demonstrate the same logic but with a component dynamic import. I assume this is in case this component import itself a lib that assumes a browser environment. However I still think this is redundant content.

garychou:
Hi @kefranabg, 

I think most people are going to run into this problem primarily when they're attempting to use a component that assumes a browser environment.  And, in that case, in addition to dynamically importing the module, they'll also need to dynamically register the component.

The current example doesn't show you how to do that.

I'm relatively new to Vue, not to mention JS, so perhaps this should be assumed knowledge. Thus, I thought the docs would be improved by showing a concise example of how to both dynamically import + register in one step.  (That, or a link to another part of the docs that show how to dynamically register a component.)

ulivz:
Actually this section want to emphasize that non-ssr-friendly components should be registered at browsers lifecycle hooks. using async component directly is a wrong solution.

ulivz:
I just updated the relevant docs at 9a8f995.

garychou:
Perfect! Thanks @ulivz!

Issue #1903, opened 2019-10-01
Cannot access page headers that are not h2 or h3


lesliecdubs:
- [x] I confirm that this is an issue rather than a question.

## Bug report
Currently, header levels are hard-coded in [VuePress core](https://github.com/vuejs/vuepress/blob/543fd6c8534cd13af5aa97d050748dd916f67ce9/packages/%40vuepress/core/lib/node/Page.js#L117). 

This limits our ability to access header levels beyond `h3`.

Our use case for this is a custom table of contents component that we are displaying on every page of our site. Our content uses `h4`s. 

We don't want header levels below `h3` to show up in our table of contents; however, we'd like for the preceding `h3` to maintain the active indicator once a user has scrolled past an `h4`.

#### What is expected?
There should be some way to customize the header levels that are stored in `this.$page.headers`.

#### What is actually happening?
Header levels stored in `this.$page.headers` are hard-coded to only include `h2` and `h3`.

cc @shortdiv

kefranabg:
Hi @lesliecdubs,

Thanks for the report.

If I correctly understand, the only thing you need is including `h4` headers data in `$page.headers` so you can display these levels on your table of contents right?

shortdiv:
Jumping in here, since @lesliecdubs and I worked on this, yeah, that would solve the problem! The other option would be pulling in the desired headers from the vuepress config.

kefranabg:
See #1730

PR #1730, opened 2019-07-30
[WIP] feat: markdown.extractHeaders option

ulivz:
**Summary**

Fix #1903

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

```js
module.exports = {
  markdown: {
    extractHeaders: ['h2', 'h3', 'h4'] // defaults to ['h2', 'h3'] 
  }
}
```

Replacement of #1111

If changing the UI of default theme, please provide the **before/after** screenshot:

**Does this PR introduce a breaking change?** (check one)

- [x] Yes
- [ ] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [x] A convincing reason for adding this feature
- [ ] Related documents have been updated
- [ ] Related tests have been updated

To avoid wasting your time, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

**This is my 1st work with [Gitpod](https://www.gitpod.io/)**

PR #1904, opened 2019-10-02
fix($core): Extract all headers tags instead of just h2, h3

shortdiv:
**Summary**
Addresses #1903

Currently, the headers are hardcoded to only include h2 and h3, this makes it hard/impossible to access any other non h2/h3 headers.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/@vuepress/core/lib/node/Page.js
+++ b/packages/@vuepress/core/lib/node/Page.js
@@ -114,7 +114,7 @@ module.exports = class Page {
         // headers
         const headers = extractHeaders(
           this._strippedContent,
-          ['h2', 'h3'],
+          ['h2', 'h3', 'h4', 'h5', 'h6'],
           markdown
         )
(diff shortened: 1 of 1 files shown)


funnycoderstar:
I have the same problem, but I think the headers are more suitable from the vuepress configuration. If not configured, the default is [h2,h3].

kefranabg:
Yep, I just noticed that there was already [an opened PR to handle this](https://github.com/vuejs/vuepress/pull/1730). The good news with this PR is that this can be configured by the user. This makes this PR obsolete I guess

kefranabg:
I'm closing this PR in favor of #1730 that will fix the issue in a better way.

PR #1977, opened 2019-10-24
fix(PWA): fixed example for the custom component

xr0master:
v-slot can only be used on components
added v-if


xr0master:
@ulivz done

PR #1972, opened 2019-10-23
fix($core): set NODE_ENV before creating app

kefranabg:
**Summary**

Fix #1967 

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

If changing the UI of default theme, please provide the **before/after** screenshot:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature
- [ ] Related documents have been updated
- [ ] Related tests have been updated

To avoid wasting your time, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

PR #2028, opened 2019-11-19
fix($core): isProd is undefined in ready hook

meteorlxy:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Currently, we set `isProd` in `app.dev()` and `app.build()`.

https://github.com/vuejs/vuepress/blob/1a930dc22e4298b648dad37fa07e2442cdcb3c10/packages/%40vuepress/core/lib/node/App.js#L491-L497

As `app.process()` runs before `app.dev()` and `app.build()`, we cannot get `isProd` in `app.process()`.

https://github.com/vuejs/vuepress/blob/1a930dc22e4298b648dad37fa07e2442cdcb3c10/packages/%40vuepress/core/lib/index.js#L18-L22

However, `ready()` hook is applied in `app.process()`, in which we may need `isProd`.

https://github.com/vuejs/vuepress/blob/1a930dc22e4298b648dad37fa07e2442cdcb3c10/packages/%40vuepress/core/lib/node/App.js#L118

So this PR fix this issue

### DIFF
--- a/packages/@vuepress/core/lib/index.js
+++ b/packages/@vuepress/core/lib/index.js
@@ -11,12 +11,14 @@ function createApp (options) {
 
 async function dev (options) {
   const app = createApp(options)
+  app.isProd = false
   await app.process()
   return app.dev()
 }
 
 async function build (options) {
   const app = createApp(options)
+  app.isProd = true
   await app.process()
   return app.build()
 }
--- a/packages/@vuepress/core/lib/node/App.js
+++ b/packages/@vuepress/core/lib/node/App.js
@@ -459,7 +459,6 @@ module.exports = class App {
    */
 
   async dev () {
-    this.isProd = false
     this.devProcess = new DevProcess(this)
     await this.devProcess.process()
     const error = await new Promise(resolve => {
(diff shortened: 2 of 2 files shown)


kefranabg:
Don't you think we should rely on the env variable? [see this PR](https://github.com/vuejs/vuepress/pull/1972/files)

meteorlxy:
@kefranabg 

I didn't notice that PR.

But I think this PR is not related to the env vars, but the context API `isProd`, which should be available in `ready()` hook. They are similar but different 🤔 .

We can rely on env vars, but we still need to set `isProd`, otherwise we are introducing a breaking change.

kefranabg:
I mean, instead of using an `isProd` property, check the `NODE_ENV`?

meteorlxy:
Yeah, that's why I said:

> We can rely on env vars, but we still need to set isProd, otherwise we are introducing a breaking change.

[ctx.isPord](https://vuepress.vuejs.org/plugin/context-api.html#ctx-isprod) is listed in the documentation. So we should be careful if we really want to deperecate it.

kefranabg:
Hmmm you're right, so maybe we could just do `this.isProd = process.NODE_ENV === 'prod'` in the `createApp` function?

meteorlxy:
close due to #1972

PR #2039, opened 2019-11-26
fix($plugin-pwa): duplicate tracking of first page (fix #2017)

### LINKED ISSUE #2017: Dual Google Analytics Pings on Initial Page Load

#### Steps to reproduce

VuePress instances including https://vuepress.vuejs.org/ are receiving duplicate Google Analytics tracking pings on initial page load

#### What is expected?

Only one "collect" ping on initial page load

Two "collect" pings are occurring on initial page load

Repro:
1. Go to https://vuepress.vuejs.org/
1. Ensure Network tab is displayed with filtering of `google-analytics`
1. Refresh the page
1. Note two, not one, tracking ping

Screen Shot 2019-11-11 at 12 00 19 PM

AjitZero:
Fixes #2017
Reasoning in [Issue](https://github.com/vuejs/vuepress/issues/2017#issuecomment-558530592).

### DIFF
--- a/packages/@vuepress/plugin-google-analytics/enhanceAppFile.js
+++ b/packages/@vuepress/plugin-google-analytics/enhanceAppFile.js
@@ -18,7 +18,6 @@ export default ({ router }) => {
 
     ga('create', GA_ID, 'auto')
     ga('set', 'anonymizeIp', true)
-    ga('send', 'pageview')
 
     router.afterEach(function (to) {
       ga('set', 'page', to.fullPath)


meteorlxy:
Thanks for your contribution, but please do not delete the PR template next time.

I found another issue, and it would be better if we can solve it, too.

## Scene 1

Open the homepage: https://deploy-preview-2039--vuepress.netlify.com/

image

Click "Get Started" (i.e. go to the guide page via vue-router):

image

__`dl` param is wrong__

## Scene 2

Open the guide page directly: https://deploy-preview-2039--vuepress.netlify.com/guide/

image

__`dl` param is correct__

## Conclusion

`dl` param is always pointing to the "previous" page.

> Notice that `dp` param is useless here according to [GA Docs](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#dp), because we don't have a `dh` param together

AjitZero:
I just checked it out. 

> `dl` param is always pointing to the "previous" page.

I tried opening other pages and switching multiple times. It's actually always pointing to the first page (whichever it may be) that we open. Only the `dp` changes at each router change.

The way it is currently written, the valid URL per route is the base URL (`/` or ) + `dp`, concatenated. I'll look into the docs and get back to you shortly.

AjitZero:
@meteorlxy This seems to be the [intended behavior](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#do_not_update_the_document_location) to track the point of entry, after which the `dp` param holds the consequent path. This is not a bug, as it is. We will need to check the GA results (dashboard) for any inconsistencies.

meteorlxy:
@AjitZero 

Hmmmm that's strange.

> For 'pageview' hits, either &dl or both &dh and &dp have to be specified for the hit to be valid

But there is no `dh` to work together.

It seems that the behavior of "analytics.js" is not consistent with the documentation of "Measurement Protocol".

Alright, let's put this issue aside and merge this PR first.

Thanks for your contribution and investigation :heart:

meteorlxy:
Oh fuck, wrong scope 😅 ....

AjitZero:
Glad I could contribute! Thank you 😇

Issue #1560, opened 2019-04-25
Vuepress building is too slow ?!


JimmyVV:
- [x] I confirm that this is an issue rather than a question.

## Bug report

My project is large, it maybe contain 6000 markdown files. Once, I meet the `dev` performance problem, and I solve it by this command `node --max_old_space_size=8192 ./node_modules/vuepress/cli.js dev src`. 

But, when I run building production, the process is too slow. I also modify the building command with `node --max_old_space_size=8192 ./node_modules/vuepress/cli.js build src`. Unfortunately, it doesn't work.

#### Version

"vuepress": "^1.0.0-alpha.47",

#### What is actually happening?

![2019-04-25 15 07 46](https://user-images.githubusercontent.com/12005455/56716092-0173b580-676c-11e9-8f60-f6ef2bb484e4.gif)

#### Other relevant information

- Your OS: MAC
- Node.js version:  8.11
- Browser version: 
- Is this a global or local install? local
- Which package manager did you use for the install?
- Does this issue occur when all plugins are disabled?

JimmyVV:
the renderer.renderToString  takes up much time. 

more than 4 seconds

<img width="674" alt="Screen Shot 2019-04-25 at 3 31 03 PM" src="https://user-images.githubusercontent.com/12005455/56717485-5107b080-676f-11e9-8692-94caf497b5be.png">

JimmyVV:
after I improve the ssr performance, the time reduce to `800+ms`

```
Rendering page: /dev/api-backend/open-api/express/by-provider/logistics.onGetQuota.htmlstart: 822.013ms
```

However, the more markdown files it compile, it will need more time.

the result is :

```
plate-message/templateMessage.getTemplateLibraryList.htmlstart: 808.080ms
Rendering page: /miniprogram/dev/api-backend/open-api/template-message/templateMessage.send.htmlstart: 789.571ms
Rendering page: /miniprogram/dev/api-backend/open-api/template-message/templateMessage.getTemplateList.htmlstart: 807.422ms
Rendering page: /miniprogram/dev/api-backend/open-api/uniform-message/uniformMessage.send.htmlstart: 802.172ms
Rendering page: /miniprogram/dev/api-backend/open-api/updatable-message/updatableMessage.createActivityId.htmlstart: 812.487ms
Rendering page: /miniprogram/dev/api-backend/summary.part.htmlstart: 768.504ms
[...]

PR #2163, opened 2020-02-02
feat($core): Improve VuePress build time

kefranabg:
**Summary**

These optimisations make sense on projects that contains a huge amount of pages to render. 

Related to #1560 

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [x] Other, please describe:
Performance enhancement

- [ ] Yes
- [x] No

**Steps:**

[Improvement #1](https://github.com/vuejs/vuepress/commit/9dce83575a8e798ee29fa8ddedacfdfbbb3e7ac6) : Render static pages in parallel instead of rendering one page after the other.
[Improvement #2](https://github.com/vuejs/vuepress/pull/2163/commits/cbe1af0b674d209beedb91c7a4692effc3118d5a) : Run extendPageData enhancers in parallel

### DIFF
--- a/packages/@vuepress/core/lib/node/Page.js
+++ b/packages/@vuepress/core/lib/node/Page.js
@@ -286,17 +286,15 @@ module.exports = class Page {
    */
 
   async enhance (enhancers) {
-    const enhancerPromises = []
-
-    for (const { name: pluginName, value: enhancer } of enhancers) {
+    const enhancerPromises = enhancers.map(({ value: enhancer, name: pluginName }) => {
       const enhancerResult = enhancer(this)
       const promise = enhancerResult instanceof Promise ? enhancerResult : Promise.resolve(enhancerResult)
 
-      enhancerPromises.push(promise.catch(e => {
+      promise.catch(e => {
         console.log(e)
         return new Error(`[${pluginName}] execute extendPageData failed.`)
-      }))
-    }
+      })
+    })
 
     const results = await Promise.all(enhancerPromises)
     const enhancerError = results.find(result => result instanceof Error)
--- a/packages/@vuepress/core/lib/node/Page.js
+++ b/packages/@vuepress/core/lib/node/Page.js
@@ -290,7 +290,7 @@ module.exports = class Page {
       const enhancerResult = enhancer(this)
       const promise = enhancerResult instanceof Promise ? enhancerResult : Promise.resolve(enhancerResult)
 
-      promise.catch(e => {
+      return promise.catch(e => {
         console.log(e)
         return new Error(`[${pluginName}] execute extendPageData failed.`)
       })
--- a/packages/@vuepress/core/lib/node/Page.js
+++ b/packages/@vuepress/core/lib/node/Page.js
@@ -286,20 +286,17 @@ module.exports = class Page {
    */
 
   async enhance (enhancers) {
-    const enhancerPromises = enhancers.map(({ value: enhancer, name: pluginName }) => {
-      const enhancerResult = enhancer(this)
-      const promise = enhancerResult instanceof Promise ? enhancerResult : Promise.resolve(enhancerResult)
-
-      return promise.catch(e => {
-        console.log(e)
-        return new Error(`[${pluginName}] execute extendPageData failed.`)
-      })
-    })
-
-    const results = await Promise.all(enhancerPromises)
-    const enhancerError = results.find(result => result instanceof Error)
-    if (enhancerError) throw enhancerError
-
-    return results
+    return Promise.all(
+      enhancers.map(
+        async ({ value: enhancer, name: pluginName }) => {
+          try {
+            await enhancer(this)
+          } catch (error) {
+            console.log(error)
+            throw new Error(`[${pluginName}] execute extendPageData failed.`)
+          }
+        }
+      )
+    )
   }
 }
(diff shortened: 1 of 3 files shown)


meteorlxy:
Tip: You can use "Create draft PR" for WIP PRs

kefranabg:
@meteorlxy Yes I forgot about this one 😅 👍

TribeWeb:
Hi @kefranabg. I have a project where the build was gradually slowing down until it fails (blows the stack) and I just tried your new code. It now renders all the pages in less than a second!

itsxallwater:
@TribeWeb I'd be curious to know how many docs in your project and if you ran this code on top of 1.2 or 1.3. I've got a project with 2k+ docs and dropping this code onto 1.3, I'm still seeing lengthy build times (15 minutes and counting at the moment). I also was getting "JavaScript heap out of memory" errors mid-build until I changed the build script to set node's `max_old_space_size` to 8192.

Keen to hear a bit more as I was working on a separate approach to pull in node's worker_threads which I had working under 1.2 and had dropped my build time from 1 hour to 7.5 minutes, though the move to 1.3 has created some issues.

TribeWeb:
> I'd be curious to know how many docs in your project and if you ran this code on top of 1.2 or 1.3

Hi @itsxallwater. It’s a much smaller project about 100 pages. I’m using it with Vuetify and moving from v1.5 to 2.2. Build time was seconds previous but now hitting out of memory errors until I tried the fix above. VuePress is 1.3. I keep hitting this problem sometimes for weird reasons. I isolated the build time problem on the previous version of Vuetify to one button and fixed by wrapping it in <ClientOnly>.

bencodezen:
Great work on this @kefranabg!

meteorlxy:
About improvement #1, https://github.com/vuejs/vuepress/pull/315#issuecomment-386339054 discuessed long time ago (in v0.x 😅). Let's check if it would break the cli output in current version.

meteorlxy:
Oh, so you removed the cli output in the PR.

Have you made some benchmark for these improvements? @kefranabg

kefranabg:
@meteorlxy I'll update the description later to add the benchmark 😄On small projects like the vuepress doc, these changes are not that effective. But on projects that have like 600pages to render, it has an interesting impact.

There is still some improvements to do to improve the build time, probably on webpack config.

kefranabg:
@meteorlxy

With more than 600 pages to render

**Without optimization:**
build time = 360 seconds
**With optimizations:**
build time = 120 seconds

packages/@vuepress/core/lib/node/Page.js
```diff
@@ -286,13 +286,20 @@ module.exports = class Page {
...
+        console.log(e)
+        return new Error(`[${pluginName}] execute extendPageData failed.`)
+      })
+    })
+
+    const results = await Promise.all(enhancerPromises)
+    const enhancerError = results.find(result => result instanceof Error)
+    if (enhancerError) throw enhancerError
+
+    return results
```
meteorlxy:
```suggestion
    return Promise.all(enhancers.map(async ({ value: enhancer, name: pluginName }) => {
      try {
        await enhancer(this)
      } catch (error) {
        console.log(error)
        throw new Error(`[${pluginName}] execute extendPageData failed.`)
      }
    }))
```

What about this?

packages/@vuepress/core/lib/node/build/index.js
```diff
@@ -89,11 +89,13 @@ module.exports = class Build extends EventEmitter {
...
     logger.wait('Rendering static HTML...')
 
-    const pagePaths = []
+    const pagePathsPromises = []
     for (const page of this.context.pages) {
-      pagePaths.push(await this.renderPage(page))
+      pagePathsPromises.push(this.renderPage(page))
     }
 
+    const pagePaths = await Promise.all(pagePathsPromises)
```
meteorlxy: From 91 to 97:

```suggestion
    const pagePaths = await Promise.all(
      this.context.pages.map(page => this.renderPage(page))
    )
```

packages/@vuepress/core/lib/node/Page.js
kefranabg: 😅 way better

Mister-Hope:
Have you checked #2221? I am afraid this change will make this issue even worse.@kefranabg

itsxallwater:
FWIW these updates took our build time down to ~10 minutes, and we're talking ~1500 docs with these plugins:

image

@Mister-Hope might be noteworthy but we're overloading the build command in the package.json to run as `node --max_old_space_size=8192 ./node_modules/vuepress/cli.js build` allowing node to use more memory.

Mister-Hope:
> FWIW these updates took our build time down to ~10 minutes, and we're talking ~1500 docs with these plugins:
> 
> image
> 
> @Mister-Hope might be noteworthy but we're overloading the build command in the package.json to run as `node --max_old_space_size=8192 ./node_modules/vuepress/cli.js build` allowing node to use more memory.
I knew node can use more memory. But I am talking about memory leak. Before changing, I don't think it's normal that the memory keep rising during build.
@itsxallwater

itsxallwater:
I don't disagree, just wanted to add context from our build experience.

Mister-Hope:
> I don't disagree, just wanted to add context from our build experience.

Let me explain my idea more carefully, I agree that the build process of pages should be parrell intead of await each other. But there must be some GC issues while building. Your FR is nice, since the offcial accepted it, I want to remind the offical that this may make the GC issue even worse.

At least I am using even more memory in 1.4.0.

itsxallwater:
I'm not arguing :)

More context for the thread at large--we've moved our build to a GitHub action and it's reliably coming in at 20 minutes.

PR #2181, opened 2020-02-13
fix($shared-utils): fix date parse logic for permalinks

enagic:
This addresses the way the date string is parsed, which is causing the input for the Date constructor to assume the date to be in ISO 8601 format (UTC). This results in the generated permalink to be one day off (depending on your local timezone).

Example:
`new Date('2020-01-01')` results in the date being parsed as `Tuesday, December 31, 2019 19:00:00 GMT-0500` for EST.

`new Date('2020-1-01')` results in the date being parsed as `Wednesday, January 01, 2020 00:00:00 GMT-0500` for EST.

As noted in [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date)

> Note: Parsing of date strings with the Date constructor (and Date.parse(), which works the same way) is strongly discouraged due to browser differences and inconsistencies.

**Summary**

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes
- [ ] No

This will change the logic used for existing permalinks, possibly offsetting the published date by 1 day.

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**


newsbielt703:
Hi, where did you meet this problem? Could you provide a repro?

Since I tested the function and saw the behavior of `new Date`, but I also found that our current code (with `getMonth` and `getDate`) somehow work fine.

enagic:
It might be difficult for you to validate the issue without resetting your clock since you might not have too much delta between your local timezone and UTC.

The Mozilla website from the link above also says...

> Support for ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01") are treated as UTC, not local.

newsbielt703:
Well, my local timezone is UTC+8, so here is what I met:

```js
const date = new Date("2020-1-1");   // I got: 2019-12-31T16:00:00.000Z
const month = date.getMonth() + 1;   //  I got: 1
const day = date.getDate(); // I got: 1
```

Obviously, the date is inconsistent as you said. However, something like `getMonth` and `getDate` looks totally fine and we are using them to get our permalink. 

That's why I wish you can link me where you met the problem.

enagic:
Oh, I got ya. I thought you were asking which branch had the issue. I don't have anything pushed in Github to demo, but I was getting the issue from version `1.3.0`, if that helps.

I'm surprised `new Date("2020-1-1")` gave you `2019-12-31`.

Here are my results:
```javascript
const d1 = new Date('2020-01-01'); // Tue Dec 31 2019 19:00:00 GMT-0500 (Eastern Standard Time)
const d2 = new Date('2020-1-1'); // Wed Jan 01 2020 00:00:00 GMT-0500 (Eastern Standard Time)

d1.getMonth() + 1; // 12
d1.getDate(); // 31

d2.getMonth() + 1; // 1
d2.getDate(); // 1
```

The effect was that, though my `.md` file was `_posts/2019-02-11/hello-world/index.md`, the permalink that was generated was `/2019-02-10-hello-world/index.html`.

kefranabg:
Hey, what should we do about this PR? @enagic if we're not able to reproduce we'll close this PR

enagic:
And here I thought, "Well it works fine in my local environment" wasn't a valid QA response!

But in all seriousness, I suspect the first step to reduce should be to reset your local clock to a different timezone. In my case, it's EST.

ludanxer:
I can indeed reproduce it after I set my timezone to EST: [repo](https://github.com/ludanxer/vuepress-support/tree/pr/2181)

packages/@vuepress/shared-utils/src/getPermalink.ts
```diff
@@ -12,6 +12,20 @@ interface PermalinkOption {
 function removeLeadingSlash (path: string) {
   return path.replace(/^\//, '')
 }
+
+function parseDateString (dateString: string): [number, number, number] {
+  const [
+    yearStr,
+    monthStr,
+    dayStr
+  ] = dateString.split('-')
```
newsbielt703: Defining date in front matter with different format will cause parser to return different type of value. For example, the type of parsed value  of `2020-02-02` is `Date` (Strictly, it's an Object), whereas `2020-2-2` will be `String`.  

The point is `split` is not in `Date`'s prototype, it cant be invoked.

enagic: I was trying to trace the code back to where that could occur... the furthest place I could tell (from reading the code statically) was from `core/lib/node/util/index.js`. There's an `inferDate` function with the following:

```
if (frontmatter.date) {
  return frontmatter.date
} else (...)
  ...
}
```

I'm guessing `frontmatter.date` could be either a string or a Date instance?

Nonetheless, I've updated the parser to support `Date`.

newsbielt703: FLY, VuePress' using `gray-matter` which depends `js-yaml` which follows standard `yaml` types, so all available formats are [here](https://yaml.org/type/timestamp.html). Thus, single digit format will not be considered as `datetime`, that's why it return a string.

newsbielt703: I checkout your last commit. It wont get error, but it still get the wrong date in EST.

enagic: If `YYYY-MM-DD` format was what was causing problems in the first place and that gets parsed into a `Date` instance, I suspect that that `gray-matter` is the root cause in that case.

I'll try to see if I can investigate further down that path.

In the meanwhile, are you able to confirm if it address the incorrect date issue when date is inferred from path name?

newsbielt703: I guess so. According to the definition:
> If the time zone is omitted, the timestamp is assumed to be specified in UTC. The time part may be omitted altogether, resulting in a date format. In such a case, the time part is assumed to be 00:00:00Z (start of day, UTC).

enagic: And I think they're correct to be unassuming and default to UTC.

I now see two paths to approach this:
- Follow the way of the crowd and force all dates to be in UTC
- Adjust to the user's local timezone by:
  1. Forcing all dates to be in UTC
  2. Offset by `Date.getTimezoneOffset()`

I think option two would make it less confusing to users, but requires more gymnastics. What do you think?

enagic: Whichever way the team decides, it won't hurt my feelings, but I've pushed the code for option 2 if that's the desired direction.

I couldn't find a straightforward way to unit test within the codebase, but I did a cheap test in my console and seems to have the desired effect:

```javascript
((date) => {
function toUtcTime (date) {
  let year = 1970
  let month = 0
  let day = 1

  if (typeof date === 'string') {
    const [
      yearStr,
      monthStr,
      dayStr
    ] = date.split('-')

    year = parseInt(yearStr, 10)
    month = parseInt(monthStr, 10) - 1
    day = parseInt(dayStr, 10)
  } else if (date instanceof Date) {
    year = date.getFullYear()
    month = date.getMonth()
    day = date.getDate()
  }

  return Date.UTC(year, month, day)
}

function addTzOffset(utc) {
  const utcDate = new Date(utc)

  return new Date(utcDate.getTime() + (utcDate.getTimezoneOffset() * 60000))
}

return new Date(addTzOffset(toUtcTime((date))))
})('2020-1-1')
```

newsbielt703: I like your approach. However, it still doesn't work fine when the param is not a string. You can use @ludanxer's [repo](https://github.com/vuejs/vuepress/pull/2181#issuecomment-590673867) to test it by defining date in 2-digits, such as `2020-01-01`.

packages/@vuepress/shared-utils/src/getPermalink.ts
```diff
@@ -12,6 +12,37 @@ interface PermalinkOption {
...
+    const [
+      yearStr,
+      monthStr,
+      dayStr
+    ] = date.split('-')
+
+    year = parseInt(yearStr, 10)
+    month = parseInt(monthStr, 10) - 1
+    day = parseInt(dayStr, 10)
+  } else if (date instanceof Date) {
```
newsbielt703: Here might be the problem. `date` is already in UTC time. There's no need to retrieve year, month, day and reassemble them.

packages/@vuepress/shared-utils/src/getPermalink.ts
```diff
@@ -33,7 +64,7 @@ export = function getPermalink ({
   }
   slug = encodeURI(slug)
 
-  const d = new Date(date)
+  const d = new Date(addTzOffset(toUtcTime(date)))
```
newsbielt703: I guess `new Date()` is  redundant.

packages/@vuepress/shared-utils/src/getPermalink.ts
enagic: There obviously must be a process to follow and setup properly in my local environment that I didn't do. I didn't think the fix would involve so many attempts so I didn't invest the time to look it up. If this final go doesn't work, I'll look into doing that.

packages/@vuepress/shared-utils/src/getPermalink.ts
enagic: Makes sense. I've updated to do just that.

packages/@vuepress/shared-utils/src/getPermalink.ts
enagic: Yes, I suppose you're right since `addTzOffset` always returns a `Date` instance.

enagic:
Do we still want to have this merged?

@newsbielt703, @ulivz, @kefranabg

PR #2225, opened 2020-03-17
feat(markdown): snippet partial import

noelmace:
Only import a portion of a code file delimited by [VS Code custom regions markers](https://code.visualstudio.com/docs/editor/codebasics#_folding), based on the region key you set via a fragment identifier (`#`):

_<project root>/docs/hello-world.md_
```md
# Hello World

<<< @/src/hello-world.js#function-def{1-2}
```

_<project root>/src/hello-world.js_:
```js
// #region function-def
function helloWorld () {
  // ..
}
// #endregion function-def

export default helloWorld
```

## Motivation

Contents presented with Vuepress (documentation, tutorials, etc.) may (often?) be related to some "real" projects. Ensuring the code snippets in the doc and the "real" code is very important for this use case. Yet, it can be difficult, or even impossible, to split each code sample in a dedicated file. Therefore, some misleading code lines may be shown in the code snippet only because the author needs it in the project.

For example, if I want to show a function in the doc, but I need an `export` in order to use this function in my project, I would have to show it in the code snippet:

```js
function foo () {
  // ...
}

export default foo;
```

## Solution

This solution add support for:

- VS Code custom region markers
  -  clearly indicate a region to import
  - **_why:_** a lot of web developers (and therefore, I guess, Vuepress users) use VS Code and should be familiar with this syntax
- fragment identifier (`path#region`)
  - indicate the name of the custom region to import
  - **_why:_** same logic as the one defined in the [URL standard](https://url.spec.whatwg.org/#concept-url-fragment) (see [W3C doc](https://www.w3.org/Addressing/URL/4_2_Fragments.html))

For more details, see the new documentation and tests.

## Compatibility

Does not introduce any breaking change: result stay exactly the same if no fragment identifier is set (no region name is set by default).
Yet, the new syntax isn't backward compatible: adding a fragment identifier to the path will lead to a "cannot find path @/path/file.ext#region" with the current version of Vuepress. 

If you want to use this feature before this PR is merged, you can use [patch-package](https://www.npmjs.com/package/patch-package) as follow: 

1. download [patches.zip](https://github.com/vuejs/vuepress/files/4342820/patches.zip)
1. extract this zip file in your project root (should create `patches/@vuepress+markdown+1.3.1.patch`)
1. add _patch-package_ to your dev dependencies:
    ```
    npm i -D patch-package
    ```
    or
    ```
    yarn add --dev patch-package postinstall-postinstall
    ```
1. add _patch-package_ in your `postinstall` script
    ```
    "scripts": {
    +  "postinstall": "patch-package"
    }
    ```

## Summary

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome 80.0.3987.132 Linux
- [x] Firefox 74 Linux
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information**

This feature is already used on a daily basis for a pretty heavy (but private) project, alongside a related Pandoc filter which provides this feature: [pandoc-import-code](https://github.com/noelmace/pandoc-import-code).

### DIFF
--- a/packages/@vuepress/markdown/__tests__/__snapshots__/snippet.spec.js.snap
+++ b/packages/@vuepress/markdown/__tests__/__snapshots__/snippet.spec.js.snap
@@ -17,6 +17,12 @@ exports[`snippet import snippet 1`] = `
 </code></pre>
 `;
 
+exports[`snippet import snippet with region and highlight 1`] = `
+<pre><code class="language-js{1,3}">function foo () {
+  // ..
+}</code></pre>
+`;
+
 exports[`snippet import snippet with highlight multiple lines 1`] = `
 <div class="highlight-lines">
   <div class="highlighted">&nbsp;</div>
@@ -34,4 +40,17 @@ exports[`snippet import snippet with highlight single line 1`] = `
 </div>export default function () {
 // ..
 }
+`;
+
+exports[`snippet import snippet with indented region 1`] = `
+<pre><code class="language-html">&lt;section&gt;
+  &lt;h1&gt;Hello World&lt;/h1&gt;
+&lt;/section&gt;
+&lt;div&gt;Lorem Ipsum&lt;/div&gt;</code></pre>
+`;
+
+exports[`snippet import snippet with region 1`] = `
+<pre><code class="language-js">function foo () {
+  // ..
+}</code></pre>
 `;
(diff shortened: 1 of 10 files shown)


kefranabg:
Hi @noelmace, thanks for opening a PR. I discussed with the team and that's definitely something we could add in the next release 👍 I'll do a review ASAP

noelmace:
Thx @kefranabg. I fixed this in the last commit and added a new test as requested.

wll8:
The @ symbol is always relative to the directory where the command is executed. And @alias cannot be used

Issue #2332, opened 2020-04-22
updateMeta removes original meta charset and viewport


sakihet:
- [x] I confirm that this is an issue rather than a question.

#### Steps to reproduce

open a page.

#### What is expected?

following meta tags which original index.html contains are shown.
```
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width,initial-scale=1">
```

#### What is actually happening?

the tags are removed by updateMeta method.

ludanxer:
Thanks for the report, introduced in #2164 .

Not only those two meta tags, it will remove all meta tags in the html template.

Janne252:
I literally started using VuePress earlier today and quickly discovered that some `meta` tags are removed at runtime. This might be common knowledge but the issue can be _temporarily_ (until the issue is resolved) fixed by re-declaring "default" meta tags in the configuration `head` section:
```js
    ...
    head: [
        ['meta', { name: 'viewport', content: 'width=device-width,initial-scale=1' }],
        ['meta', { charset: 'utf-8' }],
    ],
    ...
```

And by providing a customized `.vuepress/templates/ssr.html` template where the default `meta` tags have been removed:

```html
<!DOCTYPE html>
<html lang="{{ lang }}">
  <head>
    <title>{{ title }}</title>
    <meta name="generator" content="VuePress {{ version }}">
    {{{ userHeadTags }}}
    {{{ pageMeta }}}
    {{{ renderResourceHints() }}}
    {{{ renderStyles() }}}
  </head>
  <body>
    <!--vue-ssr-outlet-->
    {{{ renderScripts() }}}
  </body>
</html>
```
As the [documentation states](https://v1.vuepress.vuejs.org/guide/directory-structure.html#directory-structure), the `ssr.html` template should be based on the default template.

PR #2333, opened 2020-04-22
fix($core): preserve original meta charset and viewport (#2332)

sakihet:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**


sakihet:
hi @ludanxer , thanks for suggestion.

> I'd suggest we extract templateMeta from the template and unionBy them together, what do you think?

looks good. i'm going to fix.

sakihet:
i updated my PR. please review @ludanxer

packages/@vuepress/core/lib/client/root-mixins/updateMeta.js
```diff
@@ -35,10 +35,26 @@ export default {
...
+      if (typeof document !== 'undefined') {
+        const templateMetaTags = document.querySelectorAll('meta')
+        templateMetaTags.forEach(tag => {
+          const names = tag.getAttributeNames()
+          if (names.length === 1) {
+            const attr = tag.getAttributeNames()[0]
+            const obj = {}
+            obj[attr] = tag.getAttribute(attr)
+            templateMeta.push(obj)
+          } else if (names.every(e => ['name', 'content'].includes(e))) {
```
NeverBehave, line 48: wondering about the why these if statements are necessary, simply assign each attribute with their name should be fine:
e.g.
```
        const obj = {}
        for (const key of e.getAttributeNames()) {
          obj[key] = e.getAttribute(key)
        }
        return obj
```

packages/@vuepress/core/lib/client/root-mixins/updateMeta.js
```diff
@@ -35,10 +35,26 @@ export default {
 
     getMergedMetaTags () {
       const pageMeta = this.$page.frontmatter.meta || []
+      const templateMeta = []
+      if (typeof document !== 'undefined') {
+        const templateMetaTags = document.querySelectorAll('meta')
```
NeverBehave, line 40: `this.currentMetaTag` already gather meta tag on mount and reusing may provide better code readability

NeverBehave:
Wondering how this pr goes since the related issue really hurts experience on mobile devices.

packages/@vuepress/core/lib/client/root-mixins/updateMeta.js
sakihet: @NeverBehave thanks!

packages/@vuepress/core/lib/client/root-mixins/updateMeta.js
NeverBehave: > wondering about why these if statements are necessary, simply assign each attribute with their name should be fine?

e.g. https://github.com/vuejs/vuepress/pull/2360/files#diff-436014f2c7425af1edd74dfd675d069eR42

sakihet: @NeverBehave some meta tags doesn't have name attribute. then, i wrote these if statements.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

NeverBehave: I am a bit confused, I think the `getAttributeNames()` will return all names from the elements: 
```
> a
<meta charset="utf-8">

> a.getAttributeNames()
Array [ "charset" ]
```

So basically loop through each of these attributes should be fine? What exactly the name you reference to? If `<meta name=""/>` then it is also included in the function returns. 

```
> document.querySelectorAll('meta')[1]
<meta http-equiv="X-UA-Compatible" content="IE=Edge">

> document.querySelectorAll('meta')[1].getAttributeNames()
Array [ "http-equiv", "content" ]
```
The example I provided previous did pass the test tho, correct me if I miss anything

ludanxer:
@sakihet Sorry for the super delay.

After some thoughts, change how `currentMetaTags` is initialized seems to preserve all template metas.

Change

https://github.com/vuejs/vuepress/blob/d91beca8c8789624ebb82bd3736f79ef422af5f4/packages/%40vuepress/core/lib/client/root-mixins/updateMeta.js#L21

into

``` js
// init currentMetaTags from pageMeta
this.currentMetaTags = this.getMergedMetaTags().map(m => {
  let selector = 'meta'
  for (const [key, value] of Object.entries(m)) {
    selector += `[${key}="${value}"]`
  }
  return [...document.querySelectorAll(selector)]
}).flat()
```

What do you think?

packages/@vuepress/core/lib/client/root-mixins/updateMeta.js
sakihet: sorry, i misunderstood. thank you for the information.

sakihet:
@ludanxer no problem.

> After some thoughts, change how currentMetaTags is initialized seems to preserve all template metas.

looks good. thank you for your help.
i updated PR. please check again.

marchingon12:
...sooo why isn't this merged yet?

llacb47 (approved):

NeverBehave:
> ...sooo why isn't this merged yet?

Look at the reference of #2332: their own websites also affected by this bug and still nobody seems to be worried.

~~Probably Vuepress believes that nobody uses Firefox and everyone likes the wrong behavior of Chrome. Maybe fire a bug report to Firefox for following the standard?~~

PR #2338, opened 2020-04-23
fix: allows no rel attribute on external links in the nav

abea:
**Summary**

Currently it's not possible to specify that there should be no `rel` attribute on external nav links to use the default behavior. Setting to an empty string, as [indicated in the documentation](https://v1.vuepress.vuejs.org/theme/default-theme-config.html#navbar-links), is falsy, so you end up in the final ternary, which uses `noopener noreferrer` for external links.

This change allows devs to pass in a falsy, non-`undefined`, value to the `rel` property to have no rel attribute on the link. It also change the fallback of an empty string to `null` since an empty `rel` attribute has no particular meaning, so it's better to have it not there at all.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

I could not tell from the contributor guide or the package scripts how to build this for linking to a project and testing. I'd be happy to do that with some direction.
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

### DIFF
--- a/packages/@vuepress/theme-default/components/NavLink.vue
+++ b/packages/@vuepress/theme-default/components/NavLink.vue
@@ -71,7 +71,7 @@ export default {
       if (this.isNonHttpURI) {
         return null
       }
-      if (!this.item.rel && this.item.rel !== undefined) {
+      if (!this.item.rel && this.item.rel === false) {
         return null
       }
       if (this.item.rel) {
--- a/packages/@vuepress/theme-default/components/NavLink.vue
+++ b/packages/@vuepress/theme-default/components/NavLink.vue
@@ -71,10 +71,13 @@ export default {
       if (this.isNonHttpURI) {
         return null
       }
+      if (!this.item.rel && this.item.rel !== undefined) {
+        return null
+      }
       if (this.item.rel) {
         return this.item.rel
       }
-      return this.isBlankTarget ? 'noopener noreferrer' : ''
+      return this.isBlankTarget ? 'noopener noreferrer' : null
     }
   },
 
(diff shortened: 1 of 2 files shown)


packages/docs/docs/theme/default-theme-config.md
```diff
@@ -69,14 +69,14 @@ module.exports = {
 }
 ```
 
-Outbound links automatically get `target="_blank" rel="noopener noreferrer"`. You can offer `target` and `rel` to customize the attributes:
+Outbound links automatically get `target="_blank" rel="noopener noreferrer"`. You can offer `target` and `rel` to customize the attributes. Passing a "falsy" value to the `rel` attribute will disable that attribute for a link:
```
abea, line 72: Oh, maybe I should make this clearer to say that the falsy value can't be `undefined`.

newsbielt703: IMO, change the rule. Let user set `false` rather than falsy value to disable the attribute.

abea: That makes sense.

abea:
Should I create an issue to update this in the Chinese docs or something?

packages/@vuepress/theme-default/components/NavLink.vue
```diff
@@ -71,10 +71,13 @@ export default {
       if (this.isNonHttpURI) {
         return null
       }
+      if (!this.item.rel && this.item.rel === false) {
+        return null
+      }
```
newsbielt703, line 76:
```suggestion
      if (this.item.rel === false) {
        return null
      }
```

newsbielt703:
> Should I create an issue to update this in the Chinese docs or something?

I'll take it later. No worries.

PR #2380, opened 2020-05-10
feat($default-theme): allow 'auto' in defined sidebar (fix #1252)

### LINKED ISSUE #1252: Defined sidebar with auto fallback
## Feature request
Hi,
I was using ```sidebar: 'auto'``` until now.
For a specific page, I want to override the sidebar.
I thought I could so this with the in YAML config of that specific page but after looking at the docs, it looks like I need to do this in the theme config.
So I added my sidebar, but it remove the automatic sidebar from every other page !
By looking at the example for the default theme config (https://vuepress.vuejs.org/theme/default-theme-config.html#sidebar), I see there is a fallback option.
If I use the fallback options with auto, the whole site does not load but there is no error while generating.
``` 
        sidebar: {
            '/documentation/vue/': [
                '',
                'why-vue',
                'plugins',
            ],
        
            // fallback
            '/': 'auto',
        },
```
The only possible way to bypass this that works would be to not provide a fallback and add a YAML config for EVERY page with 
```
---
sidebar: auto
---
```

Allow to have an auto sidebar for everypage expect a few specific pages.

A) Allow overriding an auto sidebar in the YAML of the page
B) Allowing auto option on sidebar fallback in config.json

Option B

bobheadxi:
**Summary**

This PR allows you to define `'auto'` in defined sidebars, to allow you do define groups of pages that should have sidebars automatically generated. This fixes #1252, which has some detailed use cases + rationale from others.

The change only adds an additional case to check specifically for an `'auto'` configuration so it shouldn't be breaking

For example:

```js
// .vuepress/config.js
module.exports = {
  themeConfig: {
    sidebar: {
      '/handbook/project-management/': [
        'scope',
        'sprints',
        'repositories',
      ],
      '/handbook/tools/': [
        'slack',
        'github',
        'deployment',
      ],
      '/': 'auto',
    },
}
```

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

- [ ] Related tests have been updated (it does not seem like there is are programmatic tests for the code changed)

**Other information:**

I'm not sure how to test this - any guidance would be much appreciated! What I did do to make sure this works:

* copy the contents of `util/index.js` into my copy of vuepress in my project's `node_modules`
* verified behaviour with my project using configuration similar to the one above

### DIFF
--- a/packages/@vuepress/theme-default/util/index.js
+++ b/packages/@vuepress/theme-default/util/index.js
@@ -138,6 +138,9 @@ export function resolveSidebarItems (page, regularPath, site, localePath) {
     return []
   } else {
     const { base, config } = resolveMatchingConfig(regularPath, sidebarConfig)
+    if (config === 'auto') {
+      return resolveHeaders(page)
+    }
     return config
       ? config.map(item => resolveItem(item, pages, base))
       : []
--- a/packages/docs/docs/theme/default-theme-config.md
+++ b/packages/docs/docs/theme/default-theme-config.md
@@ -273,6 +273,8 @@ module.exports = {
         'four'   /* /bar/four.html */
       ],
 
+      '/baz/': 'auto', /* automatically generate single-page sidebars */
+
       // fallback
       '/': [
         '',        /* / */


bobheadxi:
@ulivz could you please take a look at this? :) it's a very small fix for what I feel like is significant omission from how the sidebar groups currently work, and the original issue (#1252) indicates I'm not the only one

thank you!

PR #2415, opened 2020-05-30
feat($markdown): Support for Elixir file extensions

kevinlang:
Currently, code from Elixir files is not being highlighted. This can be resolved by mapping the Elixir file endings to the name.


billyyyyy3320:
The issue is Import Code Snippets doesn't highlight Elixir code, right?

kevinlang:
yep

PR #2523, opened 2020-07-22
fix($theme-default): fix editLink for repos hosted on gitlab.com

fulopkovacs:
**Summary**
Urls that open files in the web editor follow a slightly different structure on GitLab and GitHub. This is why currently the `editLink`s in VuePress are broken for every GitLab repo.

- GitLab:
    ```
    https://gitlab.com/inkscape/inkscape/-/edit/master/.gitlab-ci.yml
    ```
- GitHub:
    ```
    https://github.com/vuejs/vuepress/edit/master/README.md
    ```

The difference is that the repository's name is followed by `/-` in the GitLab links. This PR attempts to fix this issue.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x]  Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

**Other information:**

_Tests_
Unfortunately if you attempt to edit a file on GitLab without having proper permissions, you are redirected to the blob (but there are no warning messages displayed). To properly test this PR you need a GitLab account and a GitLab repo where you have rights to editing files.

Since the only file I modified (`packages/@vuepress/theme-default/components/PageEdit.vue`) has no tests and my changes are quite small and straight-forward, I did not add new test cases to this PR.

Please let me know if you need more information, or if this PR doesn't meet the requirements! ☺️

### DIFF
--- a/packages/@vuepress/theme-default/components/PageEdit.vue
+++ b/packages/@vuepress/theme-default/components/PageEdit.vue
@@ -97,7 +97,7 @@ export default {
         : `https://github.com/${docsRepo}`
       return (
         base.replace(endingSlashRE, '')
-        + `/edit`
+        + (/gitlab.com/.test(repo) ? '/-/edit/' : '/edit')
         + `/${docsBranch}/`
         + (docsDir ? docsDir.replace(endingSlashRE, '') + '/' : '')
         + path
--- a/packages/@vuepress/theme-default/components/PageEdit.vue
+++ b/packages/@vuepress/theme-default/components/PageEdit.vue
@@ -80,8 +80,8 @@ export default {
   methods: {
     createEditLink (repo, docsRepo, docsDir, docsBranch, path) {
       const bitbucket = /bitbucket.org/
-      if (bitbucket.test(repo)) {
-        const base = outboundRE.test(docsRepo) ? docsRepo : repo
+      if (bitbucket.test(docsRepo)) {
+        const base = docsRepo
         return (
           base.replace(endingSlashRE, '')
           + `/src`
--- a/packages/@vuepress/theme-default/components/PageEdit.vue
+++ b/packages/@vuepress/theme-default/components/PageEdit.vue
@@ -92,12 +92,24 @@ export default {
         )
       }
 
+      const gitlab = /gitlab.com/
+      if (gitlab.test(docsRepo)) {
+        const base = docsRepo
+        return (
+          base.replace(endingSlashRE, '')
+          + `/-/edit`
+          + `/${docsBranch}/`
+          + (docsDir ? docsDir.replace(endingSlashRE, '') + '/' : '')
+          + path
+        )
+      }
+
       const base = outboundRE.test(docsRepo)
         ? docsRepo
         : `https://github.com/${docsRepo}`
       return (
         base.replace(endingSlashRE, '')
-        + (/gitlab.com/.test(repo) ? '/-/edit/' : '/edit')
+        + '/edit'
         + `/${docsBranch}/`
         + (docsDir ? docsDir.replace(endingSlashRE, '') + '/' : '')
         + path


fulopkovacs:
> Thanks for the PR @fulopkovacs! Do you mind refactoring this to use the pattern that we're using for `bitbucket` so we have clear scopes of different repositories?

Sure! ☺️ 

I did not follow one detail of the BitBucket implementation though: for the `base` of the _edit link_ I searched for `gitlab.com` in the value of `docsRepo`, whereas the BitBucket pattern uses the value of `repo` for some reason. As far as I know the edit links should point at the repository of the docs, right? If this is the case, then the value of `repo` is irrelevant (and `docsRepo` defaults to it when it would matter anyways).

Do you think that I have missed something? This part of the  BitBucket-style implementation looks a bit weird to me.

bencodezen:
@fulopkovacs Good question. I believe it's doing that because some sites use the actual repo for the docs while others have a mono repo that contains a separate place for docs. 

Does that help to clarify why it's implemented that way?

Definitely room for improvement, but probably should keep it consistent since I imagine it catches an edge case.

packages/@vuepress/theme-default/components/PageEdit.vue
```diff
@@ -92,12 +92,24 @@ export default {
         )
       }
 
+      const gitlab = /gitlab.com/
+      if (gitlab.test(docsRepo)) {
```
bencodezen, line 96: We should probably utilize the `repo` rather than `docsRepo` since there are two configs available to users `repo` and `docsRepo`. Once you update this, we should be good to go!

fulopkovacs: Maybe I still didn't fully understand the concept of `docsRepo`, but I thought the edit links should use that repo instead of the repo of the main site (which is the value of the `repo` variable). If it's not true, just skip the rest of my reply, and I make the change you asked for, I promise! ☺️ 

But if my understanding of `docsRepo` happens to be correct, then there is a very unique use case that my implementation handles well, but the one you suggested does not: let's suppose that you want to
- host your `repo` on GitHub (`"vuejs/vuepress"`)
- but your `docsRepo` on GitLab (`"https://gitlab.com/vuejs/vuepress"`). 

In that case, testing `repo` (`"vuejs/vuepress"`) for the `/gitlab.com/` pattern wouldn't yield a match, so the edit link would follow the GitHub-style (`"https://gitlab.com/vuejs/vuepress/edit/master/docs/guide/using-vue.md"`, note that the `-/` is missing). The reverse is also true (if `repo` is on GitLab, and `docsRepo` is on GitHub).

On the other hand, testing `docsRepo` is always safe, since it either has a value that the user explicitly set, or it falls back to the value of `repo` (see line `56` of this file).

What do you think? I think this implementation handless both the mono-repo-for-the-docs and the default situation pretty well, but if you're not sure about that, then I'll follow your suggestion without further discussion. (Sorry for taking up so much of your time, by the way, you're making an awesome framework, so you deserve some rest!)

bencodezen: You are totally right! Thanks for taking the time to explain your rationale behind your implementation.

Given that you have figured out an inefficient path in the code, would it be alright if you updated the BitBuckets implementation to match yours then? You should get full credit for this awesome refactor!

fulopkovacs: Fantastic! :) I'll update the BitBuckets-parts later today (that would've been my follow-up question anyways).

packages/@vuepress/theme-default/components/PageEdit.vue
```diff
@@ -80,8 +80,8 @@ export default {
   methods: {
     createEditLink (repo, docsRepo, docsDir, docsBranch, path) {
       const bitbucket = /bitbucket.org/
-      if (bitbucket.test(repo)) {
-        const base = outboundRE.test(docsRepo) ? docsRepo : repo
```
fulopkovacs, line 84: The second expression (`repo`) of the ternary operator would never get executed  in the refactored implementation, since the value of `docsRepo` must be a full url (it must contain the `bitbucket.org` string), which means that it will always match the `outboundRE` regex pattern (which is `/^[a-z]+:/i` (from line 4 of `packages/@vuepress/theme-default/util/index.js`)), so the `base` variable would always get its initial value from `docsRepo`. 

I tried to understand the original intention behind this line, but it just seemed to be wrong to me: if `docsRepo` had been defined by the user as `group(or user)/project` (like `vuejs/vuepress`), then the `base` variable used the value of `repo`, which is the url of the main site's repo, and not the repo of the docs. Not to mention, that a `group(or user)/project`-style `docsRepo` value suggests a `GitHub` repo and not a `BitBucket` one.

PR #2612, opened 2020-09-16
fix: Only empty the `.temp` directory at most once per run (fix #2254)

### LINKED ISSUE #2254: No such file: `node_modules/@vuepress/core/.temp/app-enhancers/global-components-2.js`

## Bug report
In development mode it sometimes breaks down when incrementally compiling the changes and I get this error in the browser console:

```
[WDS] Errors while compiling. Reload prevented.

./node_modules/@vuepress/core/.temp/app-enhancers/global-components-2.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: ENOENT: no such file or directory, open '/<path to project>/node_modules/@vuepress/core/.temp/app-enhancers/global-components-2.js'
```

The build output in the terminal does not report any errors, and will happily recompile further changes, but nothing will happen in the browser.

#### Steps to reproduce
- `vuepress dev src`
- Make some changes to files, have noticed it both when making changes to content and theme.
- It works maybe 80-90% of the time, but once the error happens, saving again doesn't fix it and i have to ctrl-c and start the development server again.

I am sorry for not providing a minimal example, but I truly have no clue which part of my code/setup is causing this. I am using a custom theme and VueX, but the issue started appearing before setting up VueX.

#### What is expected?
A normal incremental compile + hot reload.

The error pasted above.

(I am running this via WSL, which is probably why it doesn't report my browsers)

- Output of `npx vuepress info` in my VuePress project:
```
Environment Info:

  System:
    OS: Linux 4.4 Ubuntu 18.04.4 LTS (Bionic Beaver)
    CPU: (8) x64 Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz
  Binaries:
    Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
    Yarn: 1.19.1 - ~/.yarn/bin/yarn
    npm: 6.14.3 - ~/.nvm/versions/node/v10.15.0/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
  npmPackages:
    @vuepress/core:  1.4.0
    @vuepress/theme-default:  1.4.0
    vuepress: ^1.3.0 => 1.4.0
  npmGlobalPackages:
    vuepress: Not Found
```

Other than this issue, I really really like using VuePress so far, keep up the great work!

dtinth:
**Summary**

To avoid compilation errors caused by removed files.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

Ref: https://github.com/vuejs/vuepress/issues/2254#issuecomment-689457157

### DIFF
--- a/packages/@vuepress/core/lib/node/createTemp.js
+++ b/packages/@vuepress/core/lib/node/createTemp.js
@@ -3,6 +3,11 @@ const { fs, path, chalk, logger } = require('@vuepress/shared-utils')
 // Only empty the `.temp` directory at most once per run to avoid
 // compilation errors caused by removed files.
 // See: https://github.com/vuejs/vuepress/issues/2254#issuecomment-689457157
+//
+// Known issue: This can cause the `.temp` directory to grow while the server
+// is running, but the impact is limited because the `.temp` directory will
+// be cleared when restarting the server.
+// See discussion in https://github.com/vuejs/vuepress/pull/2612
 let alreadyEmptied = false
 
 /**
--- a/packages/@vuepress/core/lib/node/createTemp.js
+++ b/packages/@vuepress/core/lib/node/createTemp.js
@@ -1,5 +1,10 @@
 const { fs, path, chalk, logger } = require('@vuepress/shared-utils')
 
+// Only empty the `.temp` directory at most once per run to avoid
+// compilation errors caused by removed files.
+// See: https://github.com/vuejs/vuepress/issues/2254#issuecomment-689457157
+let alreadyEmptied = false
+
 /**
  * Create a dynamic temp utility context that allow to lanuch
  * multiple apps with isolated context at the same time.
@@ -19,8 +24,9 @@ module.exports = function createTemp (tempPath) {
 
   if (!fs.existsSync(tempPath)) {
     fs.ensureDirSync(tempPath)
-  } else {
+  } else if (!alreadyEmptied) {
     fs.emptyDirSync(tempPath)
+    alreadyEmptied = true
   }
 
   logger.debug(`Temp directory: ${chalk.gray(tempPath)}`)


billyyyyy3320 (commented):
Thanks! Sorry for the very late review.

I'll have to check again how we use `.temp`. Before that, how would you suggest us to test it? I couldn't reproduce the error. Do you have a repo for reproduce?

dtinth:
Thanks for reviewing!

I haven’t dug deep into why it happens, but it seems like a race condition when recompiling. When the directory is emptied, sometimes it causes problem with the file watcher (it stoppped watching files).

I will create a minimal reproducable case and update you.

dtinth:
@billyyyyy3320 Here is a reproduction case.

https://codesandbox.io/s/nostalgic-meadow-456zx?file=/README.md

I put it on CodeSandbox so that we are reproducing the issue on the same environment. I made sure to use the latest version of VuePress.

billyyyyy3320 (commented):
According to the commit message https://github.com/vuejs/vuepress/commit/b507faeaf86f48c45e3a7a999b7fa7c058b82d00 , The code changes was trying to empty temp dir before launching app instead of before reloading app.

However, I found [plugin-register-components](https://github.com/vuejs/vuepress/blob/master/packages/%40vuepress/plugin-register-components/index.js#L55) generates new files whenever app reload. It means there'll be redundant files in temp dir, and they'll keep increasing until the dev server restart.

It'd be great if you could help fix the issue with plugin-register-components. Or please at least leave a comment about this known issue in codebase or github issue (IMO, the original issue you fixed is important than this).

dtinth:
@billyyyyy3320 Yes, this is an expected side effect, but even with that, the dev server no longer crashes. This solution is backed by hours of real world usage.

It has been shown that emptying the temp directory was not safe while the server is running. I could not figure out how to make it safe quickly enough so I settled on this solution for now.

This solution is not perfect, but in my opinion, I would prefer a bloated temp directory than a server that crashes every 10 minutes of usage.

In the [original issue, a suggested fix that works was to disable that line entirely](https://github.com/vuejs/vuepress/issues/2254#issuecomment-658944622). I also worry about temp directory getting too big, so I opted for a compromise. Instead of not emptying the directory at all, at least empty it when starting server.

billyyyyy3320 (approved):

dtinth:
Thanks @billyyyyy3320!

Issue #2614, opened 2020-09-18
Can not have multiple meta "article:tag"


taichunmin:
- [x] I confirm that this is an issue rather than a question.

#### Steps to reproduce

Write multiple meta article:tag in frontmatter

```yaml
meta:
  - property: article:tag
    content: tag1
  - property: article:tag
    content: tag2
  - property: article:tag
    content: tag3
```

#### What is expected?

```html
<meta property="article:tag" content="tag1">
<meta property="article:tag" content="tag2">
<meta property="article:tag" content="tag3">
```

#### What is actually happening?

```html
<meta property="article:tag" content="tag1">
```

#### Other relevant information

- Output of `npx vuepress info` in my VuePress project:

https://github.com/vuejs/vuepress/blob/3c94f711434ae7fd5f983ca19c8a2c29ef22e855/packages/%40vuepress/core/lib/client/root-mixins/updateMeta.js#L36

Need to patch `getMergedMetaTags()`

Issue #2607, opened 2020-09-11
Desktop view while on mobile browser


treagod:
- [x] I confirm that this is an issue rather than a question.

## Bug report

When using the Firefox browser on Android my Vuepress application does not look friendly. It looks exactly like the desktop version

#### Steps to reproduce

Visit [Vuepress Website](https://vuepress.vuejs.org/) or visit your own generated Vuepress application (`yarn create vuepress`) via Firefox Mobile (it also affects Firefox Desktop with developer mobile view activated)

#### What is expected?

See a mobile view of the vuepress application, i.e. a mobile navigation with hamburger icon.

#### What is actually happening?

See the desktop view without a hamburger icon and all navigation links still visible. Also the content is very small on the sccreen.

#### Other relevant information

- Mobile: Firefox 80.0.1 on Android
- Desktop: Firefox 80.0 on Ubuntu

billyyyyy3320:
Tried to reproduce by using Browerstack. 

It looks good on:
- Firefox 80 on Win10(Windows), macOs Catalina(macOs), iPhone11(iOS), Samsung Galaxy S20 and Google Pixel 4(Android)

treagod:
That's strange. I made a screenshot, don't know if that helps or if I can do anything else. 

![Screenshot_2020-09-12-10-35-04-784_org mozilla firefox](https://user-images.githubusercontent.com/4701538/92991361-a6b73700-f4e3-11ea-954c-cac2dae98553.jpg)

ktquez:
@billyyyyy3320 

In the source code, the following meta tags are present:
```html
<meta charset="utf-8">
<meta name="viewport" content="width=device-width,initial-scale=1">
```

However, when the site is loaded they are removed.
If you "inspect" you will see that these 2 tags disappear.

I did the test here (Android + Firefox), and the site loads normally with the mobile layout, but after a few seconds it changes to the desktop version.

amzrk2:
Basically same issue here.

When I inspect the site on Firefox 82.1.3 on Android it rendered normally with the mobile layout on first load, then turned to desktop layout instantly.

For now, as @ktquez said, I add these two tags in `docs/.vuepress/config.js` same as injecting extra tags to `<head>` which can fix this temporarily.

However, these two tags also disappears after I try inspecting the site with latest Chrome on Android. So I guess that it's just new Firefox's (after Daylight update) default viewport setting that differs from Chrome which causes this issue.

[...]

PR #2615, opened 2020-09-18
fix meta type array #2614

taichunmin:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**


seifertm:
Great work! I just ran into the same issue and was happy to see that someone already addressed it :)

I think that your PR only addresses a part of the problem, though. For example the `og:image` tag can have multiple values and is not included in `META_TYPE_ARRAY`. This opens a can of worms, because each `og:image` tag [can have properties](https://ogp.me/#array) like `og:image:width` and `og:image:height` which would also need to be included.

I see two solutions for this:
1. Make the list exhaustive. This is not a problem for the Open Graph specification, but Facebook, Twitter and the sorts use their own meta tags. Including all of them in the list and maintaining the list of tags might be a pain

2. Instead of using a list of meta tags that can occur multiple times, find a smarter way than `lodash.unionBy` to merge the page and site meta tags. Do you have any ideas how we could do that?

taichunmin:
I have no idea of how to allow the meta type array without whitelist, maybe we can expose the whitelist as a vuepress config?

taichunmin:
I found another problem: the order of meta tags is very important.

For example:

```xml
<meta property="og:image" content="https://example.com/rock.jpg" />
<meta property="og:image:width" content="300" />
<meta property="og:image:height" content="300" />
<meta property="og:image" content="https://example.com/rock2.jpg" />
<meta property="og:image" content="https://example.com/rock3.jpg" />
<meta property="og:image:height" content="1000" />
```

means there are 3 images on this page, the first image is 300x300, the middle one has unspecified dimensions, and the last one is 1000px tall.

taichunmin:
Maybe we should maintain a list for unique meta name?   We can only deduplicate meta tags which name in the list, and we can expose the list to vueconfig?

taichunmin:
Another method is merge `siteMeta` only if `pageMeta` didn't have any of same meta name?

seifertm:
Good point with the order of meta tags!

> Another method is merging `siteMeta` only if `pageMeta` didn't have any of same meta name?

I think this would be the best solution. That way, we don't have to maintain or expose the META_TYPE array. It also means that users can provide default values in `siteMeta` and override them on a page-by-page basis in `pageMeta`. I really like the idea!

That's just my personal opinion, though. It would be nice to hear what the maintainers think :)

taichunmin:
@seifertm Okay, I will rewrite my PR to use this method.

xr0master:
Hey guys. Will it fix this issue?
https://github.com/vuejs/vuepress/issues/2607

Issue #2627, opened 2020-09-25
Duplicate ID "outbound-link-title" violates HTML5


katelynsills:
- [x] I confirm that this is an issue rather than a question.

#### Steps to reproduce

1. Create a site with several external links on one page.
2. Build the site with Vuepress 1.6.0
3. Attempt to validate the result as HTML5
4. Validation will fail due to duplicate IDs whereas with Vuepress 1.5, it did not.

For a concrete example, please see [this PR which attempts to upgrade to Vuepress 1.6.0](https://github.com/Agoric/documentation/pull/318/checks?check_run_id=1168010572) and uses a Github Action for HTML5 Validation.

#### What is expected?

Vuepress should be HTML5 compliant.

#### What is actually happening?

Duplicate IDs per page are being produced

#### Other relevant information

https://github.com/vuejs/vuepress/commit/8d101192623e0df903c56500b2cde24125424fe0 Seems to be the commit that introduced this problem.

- Output of `npx vuepress info` in my VuePress project:
```shell
Environment Info:

  System:
    OS: macOS 10.15.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  Binaries:
    Node: 14.11.0 - ~/.nvm/versions/node/v14.11.0/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.11.0/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Edge: Not Found
    Firefox: Not Found
    Safari: 14.0
  npmPackages:
    @vuepress/core:  1.6.0 
    @vuepress/theme-default:  1.6.0 
    vuepress: ^1.6.0 => 1.6.0 
  npmGlobalPackages:
    vuepress: Not Found
```

pbek:
Thank you for bringing that up, @katelynsills. I just found out about VuePress today and the first thing I did after a `yarn run build` on the boilerplate webpage was uploading the html files to <https://validator.w3.org/nu/#file> to check the HTML5 validity. 😁

adico1:
I can take that.
do you think it will be a good solution to just remove the id. I didn't find its being used.

adico1:
1. removing the id remove the first error.

but brings up another one
[...]

PR #2650, opened 2020-10-09
fix($core): #2627 Dup ID violates HTML5-4

adico1:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

- [x] Chrome - Version 86.0.4240.75 (Official Build) (x86_64)
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

### DIFF
--- a/packages/@vuepress/core/lib/client/components/OutboundLink.vue
+++ b/packages/@vuepress/core/lib/client/components/OutboundLink.vue
@@ -1,35 +1,34 @@
 <template>
-  <svg
-    class="icon outbound"
-    xmlns="http://www.w3.org/2000/svg"
-    :aria-labelledby="rIdx"
-    role="img"
-    focusable="false"
-    x="0px"
-    y="0px"
-    viewBox="0 0 100 100"
-    width="15"
-    height="15"
+  <span
+    :title="openInNewWindowTitle"
+    class="sr-only"
   >
-    <title :id="rIdx">{{ openInNewWindowTitle }}</title>
-    <path
-      fill="currentColor"
-      d="M18.8,85.1h56l0,0c2.2,0,4-1.8,4-4v-32h-8v28h-48v-48h28v-8h-32l0,0c-2.2,0-4,1.8-4,4v56C14.8,83.3,16.6,85.1,18.8,85.1z"
-    />
-    <polygon
-      fill="currentColor"
-      points="45.7,48.7 51.3,54.3 77.2,28.5 77.2,37.2 85.2,37.2 85.2,14.9 62.8,14.9 62.8,22.9 71.5,22.9"
-    />
-  </svg>
+    <svg
+      class="icon outbound"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-hidden="true"
+      role="img"
+      focusable="false"
+      x="0px"
+      y="0px"
+      viewBox="0 0 100 100"
+      width="15"
+      height="15"
+    >
+      <path
+        fill="currentColor"
+        d="M18.8,85.1h56l0,0c2.2,0,4-1.8,4-4v-32h-8v28h-48v-48h28v-8h-32l0,0c-2.2,0-4,1.8-4,4v56C14.8,83.3,16.6,85.1,18.8,85.1z"
+      />
+      <polygon
+        fill="currentColor"
+        points="45.7,48.7 51.3,54.3 77.2,28.5 77.2,37.2 85.2,37.2 85.2,14.9 62.8,14.9 62.8,22.9 71.5,22.9"
+      />
+    </svg>
+  </span>
 </template>
 
 <script>
 export default {
-  data: function () {
-    return {
-      rIdx: 'outbound-link-title-' + Math.floor(Math.random() * 1000000)
-    }
-  },
   computed: {
     openInNewWindowTitle () {
       return this.$themeLocaleConfig.openNewWindowText || '(opens new window)'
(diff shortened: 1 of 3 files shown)


packages/@vuepress/core/lib/client/components/OutboundLink.vue
```diff
@@ -1,26 +1,30 @@
...
-    <polygon
-      fill="currentColor"
-      points="45.7,48.7 51.3,54.3 77.2,28.5 77.2,37.2 85.2,37.2 85.2,14.9 62.8,14.9 62.8,22.9 71.5,22.9"
-    />
-  </svg>
+    <svg
+      class="icon outbound"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-hidden="true"
+      role="img"
```
ktquez, line 10: I think you can remove `role="img"`

packages/@vuepress/core/lib/client/components/OutboundLink.vue
```diff
@@ -1,26 +1,30 @@
...
-    role="img"
-    focusable="false"
-    x="0px"
-    y="0px"
-    viewBox="0 0 100 100"
-    width="15"
-    height="15"
+  <span
+    :title="openInNewWindowTitle"
+    class="sr-only"
```
ktquez, line 4: In this case, I think native text increases the chances of support in combinations of screen readers and browsers.

```html
<span>
  <span class="sr-only">{{ openInNewWindowTitle }}</span>
```

ktquez:
We can use the `sr-only` class used by [TailwindCSS](https://tailwindcss.com/docs/screen-readers#class-reference) and put it in some global vuepress style file.

```css
.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  white-space: nowrap;
  border-width: 0;
}
```

packages/@vuepress/core/lib/client/components/OutboundLink.vue
adico1: thank you, its my first time contributing so I missed your remarks just saw that now.

adico1: you said "I think", so I went browsing a bit:
MDN say nothing special about it but they have a side note about
**Deciding between aria-hidden="true", role="presentation", and role="none"**
Which means that to make sr ignore the SVG is should use role that is not 'img', therefore I think we can conclude that it's unnecessary.

Also the next article I references "Contextually Marking up accessible images and SVGs" doesn't use role with aria-hidden. 

So I'll remove it. 
reference:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute
https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html

packages/@vuepress/core/lib/client/components/OutboundLink.vue
adico1: Just to be sure I understand you:
```
<span>
  <svg
        class="icon outbound"
        xmlns="http://www.w3.org/2000/svg"
        aria-hidden="true"
        role="img"
        focusable="false"
        x="0px"
        y="0px"
        viewBox="0 0 100 100"
        width="15"
        height="15"
      >
        <path
          fill="currentColor"
          d="M18.8,85.1h56l0,0c2.2,0,4-1.8,4-4v-32h-8v28h-48v-48h28v-8h-32l0,0c-2.2,0-4,1.8-4,4v56C14.8,83.3,16.6,85.1,18.8,85.1z" />
        <polygon fill="currentColor"
          points="45.7,48.7 51.3,54.3 77.2,28.5 77.2,37.2 85.2,37.2 85.2,14.9 62.8,14.9 62.8,22.9 71.5,22.9" />
  </svg>
  <span :title="openInNewWindowTitle" class="sr-only"></span>
</span>```

adico1: As shown in the code snippet above, I incorporated all your comments:
- [X] added a wrapping span for the SVG and sr-span
- [X] moved the parent sr-span to be a sibling (rather then parent of the SVG)
- [X] made the sr-info text as text child of the sr-span instead of title attribute

packages/@vuepress/core/lib/client/components/OutboundLink.vue
ktquez: Sorry, I said "I think" because `role="img"` is ignored when `aria-hidden` is present, so removing it or leaving it wouldn't make much difference.

You did well to remove it.

ktquez (approved):

bencodezen (approved):
Thanks for your work on this @adico1!

Mister-Hope:
image
Is it bugs? Or design to be like this? @adico1 @bencodezen @ktquez

Mister-Hope:
Hey, I found the problems, it's a internal plugin, why are you placing styles in theme-default? It should be placed under Outboundlink.

**Otherwise you are breking every theme which is not extending @vue/default!!!**

adico1:
I can fix that, what is the proper location for the style?

Issue #2153, opened 2020-01-26
Set Metadata on component level


samanthaming:
## Feature request

Currently the way to set custom metadata is through the frontmatter in the markdown file. It'd be very helpful if we could set this on the component level.

### The Problem

Here's a scenario of why this is important. I use Cloudinary to store my images and I use a gulp task to upload the images. Which then populates a single javascript object with all of the ID of the images. These are the images I want to use in my meta head (especially for my social media links on twitter and facebook). And I really don't want to set them manually in each of my markdown frontmatter. 

### Ideal Solution

It'd be so nice, if I could set the meta head directly in my layout or component. Something like [vue-meta](https://github.com/nuxt/vue-meta) from Nuxt.

### Alternative Solution

Unfortunately, I had trouble trying to figure out how to set the `vue-meta` plugin to work with VuePress via the SSR way. So instead I created a custom plugin that will programmatically set the frontmatter with `frontmatter.meta`. It's working in my dev environment, however, when the site is built, it then has duplicates (I just found this out and have not looked into it yet, when I find out why, I'll update this issue).

### Conclusion

I know VuePress is mainly built for technical documentation. But for those using it for more than that, I think meta data is a super important feature.  I used VuePress to build my [blog](https://www.samanthaming.com/) and I picked a static site generator for SEO reason. And meta data is an integral part of that.

Overall I loved using VuePress to build my new site (the markdown support and the built-in header search are amazing!). All that's missing is this meta data support. Not trying to compare VuePress to others, but Gridsome offers this. And it's probably one of the feature that might make me want to switch over.

Anyways, I think the team did a fabulous job on this. I hope you will consider this feature! 👏

bencodezen:
Thanks for filing this @samanthaming! We really appreciate the time and effort you took to investigate this and will add this our list of items to discuss for our roadmap! Would you be interested in sharing your current plugin implementation so we can figure out a best path forward?

samanthaming:
Thanks @bencodezen for looking into this 👏  Absolutely! I'll reach out to you and we can set something up 🙌

adamdehaven:
To follow up on @samanthaming, I also wrote a custom plugin that injects metadata into the `ssrTemplate` - would be glad to share, if helpful.

samanthaming:
@adamdehaven wooo you have a custom plugin 🤩 is it an open source that you can provide a link to? Would love to check it out 😊

adamdehaven:
[...]

PR #2658, opened 2020-10-12
feat($core): add canonical link to frontmatter

d-pollard:
**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

**Other information:**

This feature introduces the ability to specify `canonical` in the frontmatter section on a page:

```
---
canonical: https://google.com
---
```

Which will then render the proper head HTML tag, and also keep it updated when the page changes

### DIFF
--- a/packages/@vuepress/core/lib/client/index.ssr.html
+++ b/packages/@vuepress/core/lib/client/index.ssr.html
@@ -7,7 +7,7 @@
     <meta name="generator" content="VuePress {{ version }}">
     {{{ userHeadTags }}}
     {{{ pageMeta }}}
-    {{{ canonical() }}}
+    {{{ canonical }}}
     {{{ renderResourceHints() }}}
     {{{ renderStyles() }}}
   </head>
--- a/packages/@vuepress/core/lib/client/index.ssr.html
+++ b/packages/@vuepress/core/lib/client/index.ssr.html
@@ -7,6 +7,7 @@
     <meta name="generator" content="VuePress {{ version }}">
     {{{ userHeadTags }}}
     {{{ pageMeta }}}
+    {{{ canonical() }}}
     {{{ renderResourceHints() }}}
     {{{ renderStyles() }}}
   </head>
--- a/packages/@vuepress/core/lib/client/index.ssr.html
+++ b/packages/@vuepress/core/lib/client/index.ssr.html
@@ -7,7 +7,7 @@
     <meta name="generator" content="VuePress {{ version }}">
     {{{ userHeadTags }}}
     {{{ pageMeta }}}
-    {{{ canonical }}}
+    {{{ canonicalLink }}}
     {{{ renderResourceHints() }}}
     {{{ renderStyles() }}}
   </head>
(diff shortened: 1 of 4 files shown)


d-pollard:
Partial resolution to https://github.com/vuejs/vuepress/issues/2153

adamdehaven:
@d-pollard I can confirm (with your latest addition) that the `frontmatter.canonical` property works, and is also exposed to the `extendPageData` Options API, meaning it can be dynamically injected into the `frontmatter` :tada: 

This should **definitely** roll out ASAP - you're awesome!

adamdehaven:
@d-pollard don't forget: the docs would need to be updated for this PR as well. 

Now you've got me thinking about injecting structured data in a similar way. It also would need to be updated on each route change. I wonder if I'm the only person that would like a hook in the Options API that would allow for injecting (via `extendPageData`) custom content/tags. In my use-case, I'd want to inject a `<script>` tag with JSON into the `<head>`, something like this:

``` js 
const structuredData = {
    '@context': 'https://schema.org/',
    '@type': 'WebSite',
    name: this.meta_data.title + (this.$page.frontmatter.subtitle ? ' | ' + this.$page.frontmatter.subtitle : '') || null,
    description: this.meta_data.description || null,
    url: frontmatter.canonical,
    image: {
        '@type': 'ImageObject',
        url: this.$site.themeConfig.domain + this.$site.themeConfig.defaultImage,
    },
    '@id': frontmatter.canonical,
}
```

``` html
<script type="application/ld+json">{{ JSON.stringify(structuredData, null, 4) }}</script>
```

I'm currently doing this with a custom component; however, it is technically rendered after the DOM, meaning it isn't available when the page is scraped (before JavaScript is enabled).

d-pollard:
Yeah we could definitely look into a custom hook

Typically I write the feature in it's entirety then open another PR with docs code so that if the version changes, the feature code is still introduced and I can update the docs to reflect the proper version

d-pollard:
Like in this case, I have reservations about using just the keyword canonical so with probably change it to be more idiomatic (thinking `canonicalLink` to reflect current codebase practices)

adamdehaven:
@d-pollard I almost suggested a change regarding the `canonical` keyword. I post content on `dev.to` and they actually use the YAML keyword `canonical_url` instead, which, I think would be good; however, since other keywords are camelCase, maybe `canonicalUrl` instead? 

I lean towards `url` rather than `link` because `link` seems to reference more of the HTML `<link>` element itself, whereas url references the value, which is what the user would be providing in the `frontmatter`.

d-pollard:
Yeah that makes sense, I'll make the changes probably tomorrow, and then document the feature once it's all merged in.


Doing this makes me wonder if I could also use the same pattern to introduce vue-meta in a way that's backwards compatible

d-pollard:
I fibbed; I've updated the PR with the new terms; used `canonicalLink` for the SSR injection to reflect the HTML element it's injecting, and I've used `canonicalUrl` for the frontmatter value to reflect the actual use case

adamdehaven:
That all looks good to me.

bencodezen (approved):
Thanks for the collab effort @d-pollard and @adamdehaven!

PR #2662, opened 2020-10-14
fix($core): add missing styles for `<OutboundLink />`

Mister-Hope:
image

#2627 adds a hint with `sr-only` class, but the class style is placed in `@vue/theme-default` by mistake.

OutboundLink is an internal component, and will effect all the themes, so the style must be placed in the component.

### DIFF
--- a/packages/@vuepress/core/lib/client/components/OutboundLink.vue
+++ b/packages/@vuepress/core/lib/client/components/OutboundLink.vue
@@ -43,3 +43,16 @@ export default {
   top: -1px;
 }
 </style>
+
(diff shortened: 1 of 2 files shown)


billyyyyy3320 (approved):

Mister-Hope:
Could you please publish 1.7.1 for this? Because it's affecting all third party theme which is not using theme extending and extend the default theme?

meteorlxy:
Not all theme is using stylus, so it's not a good idea to write stylus code in core package.

See #1482

@Mister-Hope

PR #2666, opened 2020-10-15
fix($core): reference correct canonical Url frontmatter property (fix #2665)

### LINKED ISSUE #2665: New `canonicalUrl` frontmatter property is incorrectly mapped in `ClientComputedMixin.js`

#### What is expected?

Canonical Url is grabbed from the `canonicalUrl` entry in the `frontmatter`

The canonical URL is not being injected into the SSR template because in PR #2658 the frontmatter entry was renamed (during development) **to** `canonicalUrl` **from** `canonical`. 

Line `69` in `packages/@vuepress/core/lib/node/ClientComputedMixin.js` is looking for a non-existent property called `canonical` and needs to be updated to the following:

``` js
const { canonicalUrl } = this.$page.frontmatter
```

Tagging @d-pollard

adamdehaven:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge
- [ ] IE

### DIFF
--- a/packages/@vuepress/core/lib/node/ClientComputedMixin.js
+++ b/packages/@vuepress/core/lib/node/ClientComputedMixin.js
@@ -66,10 +66,10 @@ module.exports = siteData => {
     }
 
     get $canonicalUrl () {
-      const { canonical } = this.$page.frontmatter
+      const { canonicalUrl } = this.$page.frontmatter
 
-      if (typeof canonical === 'string') {
-        return canonical
+      if (typeof canonicalUrl === 'string') {
+        return canonicalUrl
       }
 
       return false


d-pollard (approved):
Good catch! LGTM

billyyyyy3320 (approved):
LGTM!

adamdehaven:
@d-pollard curious, does this replace the code in the current `1.7.0` or will it be in a future patch? (and if future, when?)

d-pollard:
It'll probably be a different patch, so we'll need to update the docs once the release. Will try to get it in ASAP

adamdehaven:
Ah ok - the docs are correct in the `frontmatter.canonicalUrl` reference. The only thing that would need to be updated is the version number 👍

d-pollard:
@adamdehaven - this should be out now with version 1.7.1

d-pollard:
docs updated too 👍

PR #2454, opened 2020-06-19
fix($core): render static html via stream pipe (fix #1819)

sgtcoolguy:
**Summary**
This PR attempts to handle larger doc sites to avoid blowing up the NodeJs heap as seen in #1819, as well as provide a better experience by not just silently chewing through thousands of files. The PR attempts to achieve this by:
- Using [p-limit](p-limit) to limit the number of static HTML pages it attempts to render simultaneously. I arbitrarily chose 50.
- Using `bundleRenderer.renderToStream()` and then using `fs.writeStream()` and `stream.pipe()`. This avoids rendering to a string that we then write to file. It *should* pipe it to the file as it goes (which I assume should reduce RAM usage)
- Providing a nice progress bar that looks/feels slightly similar to the Webpack client./server ones earlier in the process.

Screen Shot 2020-06-19 at 10 05 55 AM

I should note that for our site with 1618 pages, I still had to increase the Node heap or it would sometimes blow up during the Webpack client/server bundling.

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [X] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [X] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [X] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature
- [ ] Related documents have been updated
- [ ] Related tests have been updated

To avoid wasting your time, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

While this adds an explicit dependency on p-limit, I did find it in the `node_modules` tree of my site, and I believe it was already deeply nested in the dependency chain of VuePress: `@vuepress/core > copy-webpack-plugin > p-limit` I aligned versions so it used `^2.2.0`.

PR #2953, opened 2021-11-24
feat: add build concurrency control (fix #1819)

### LINKED ISSUE #1819: FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
- [ x] I confirm that this is an issue rather than a question.

#### What is expected?
Normal build

```
sh deploy.sh
$ npx --max_old_space_size=8192 vuepress build docs
wait Extracting site metadata...

.....

Language does not exist tip
wait Rendering static HTML...
Rendering page: /elasticsearch-senior/depth-search/11-dis-max-best-fields.html
<--- Last few GCs --->

[8980:000002394C5333A0]   373053 ms: Mark-sweep 1368.8 (1442.2) -> 1368.8 (1442.2) MB, 1347.0 / 0.0 ms  allocation failure GC in old sp
ace requested
[8980:000002394C5333A0]   374704 ms: Mark-sweep 1368.8 (1442.2) -> 1368.8 (1427.2) MB, 1574.2 / 0.0 ms  last resort GC in old space req
uested
[8980:000002394C5333A0]   376188 ms: Mark-sweep 1368.8 (1427.2) -> 1368.8 (1427.2) MB, 1484.7 / 0.0 ms  last resort GC in old space req
uested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0000019575F25879 <JSObject>
    2: exec(this=000000FD275E10B1 <JSRegExp <String[40]: ^\/cache-pdp\/dr\/131\.html(?:\/(?=$))?$>>,0000026741B33FA9 <String[62]: /elas
ticsearch-senior/depth-search/11-dis-max-best-fields.html>)
    3: 0000008D3FABD9D1 <Symbol: Symbol.match>(aka [Symbol.match])(this=000000FD275E10B1 <JSRegExp <String[40]: ^\/cache-pdp\/dr\/131\.
html(?:\/(?=$))?$>>,0000026741B33FA9 <String[62]: /elasticsearch-s...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node_module_register
 2: v8::internal::FatalProcessOutOfMemory
 3: v8::internal::FatalProcessOutOfMemory
 4: v8::internal::Factory::NewCodeRaw
 5: v8::internal::Factory::NewCode
 6: v8::internal::compiler::Linkage::GetSimplifiedCDescriptor
 7: v8::internal::RegExpImpl::Exec
 8: v8::internal::RegExpImpl::Exec
 9: v8::internal::RegExpImpl::Exec
10: v8::internal::RegExpImpl::Exec
11: v8::internal::RegExpImpl::Exec
12: v8::internal::RegExpImpl::Exec
13: v8::internal::interpreter::BytecodeArrayRandomIterator::UpdateOffsetFromIndex
14: 0000029C5CD843C1
error Command failed with exit code 3.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

An error occurred while building the repository：https://github.com/zq99299/note-book

troyeagle:
**Summary**

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

This PR is all about build phase. No need to check browser compability.

**Other information:**

To fix #1819. Vuepress build would fail for many docs, for Promise.all push thousands of files to VueServer to render concurrently and cause OOM. 
#2454 tried to fix this problem. However, with pLimit fixed to 50, the build process may be significantly slowed down. To fix this, this PR makes it optional as a `vuepress build` param.

### DIFF
--- a/packages/@vuepress/core/lib/node/build/index.js
+++ b/packages/@vuepress/core/lib/node/build/index.js
@@ -95,7 +95,6 @@ module.exports = class Build extends EventEmitter {
 
     const pagePaths = []
     while (this.context.pages.length) {
-      // 100 at a time
       const segmentPaths = await Promise.all(
         this.context.pages.splice(0, this.maxConcurrency || 100000)
         .map(page => this.renderPage(page))
--- a/packages/@vuepress/core/lib/node/build/index.js
+++ b/packages/@vuepress/core/lib/node/build/index.js
@@ -20,6 +20,7 @@ module.exports = class Build extends EventEmitter {
   constructor (context) {
     super()
     this.context = context
+    this.maxConcurrency = this.context.options.maxConcurrency
     this.outDir = this.context.outDir
   }
 
@@ -90,10 +91,17 @@ module.exports = class Build extends EventEmitter {
 
     // render pages
     logger.wait('Rendering static HTML...')
+    if (this.maxConcurrency) logger.info(`max concurrency set: ${this.maxConcurrency}`)
 
-    const pagePaths = await Promise.all(
-      this.context.pages.map(page => this.renderPage(page))
-    )
+    const pagePaths = []
+    while (this.context.pages.length) {
+      // 100 at a time
+      const segmentPaths = await Promise.all(
+        this.context.pages.splice(0, this.maxConcurrency || 100000)
+        .map(page => this.renderPage(page))
+      )
+      pagePaths.push(...segmentPaths)
+    }
 
     readline.clearLine(process.stdout, 0)
     readline.cursorTo(process.stdout, 0)
--- a/packages/@vuepress/core/lib/node/build/index.js
+++ b/packages/@vuepress/core/lib/node/build/index.js
@@ -91,12 +91,15 @@ module.exports = class Build extends EventEmitter {
 
     // render pages
     logger.wait('Rendering static HTML...')
-    if (this.maxConcurrency) logger.info(`max concurrency set: ${this.maxConcurrency}`)
 
+    // If maxConcurrency is set, instead of rendering all pages concurrently,
+    // build task would render pages by smaller group to prevent OOM.
+    if (this.maxConcurrency) logger.info(`max concurrency set: ${this.maxConcurrency}`)
     const pagePaths = []
-    while (this.context.pages.length) {
+    const maxConcurrency = this.maxConcurrency || 100000
+    for (let i = 0; i < this.context.pages.length; i += maxConcurrency) {
       const segmentPaths = await Promise.all(
-        this.context.pages.splice(0, this.maxConcurrency || 100000)
+        this.context.pages.slice(i, i + maxConcurrency)
         .map(page => this.renderPage(page))
       )
       pagePaths.push(...segmentPaths)
--- a/packages/docs/docs/api/cli.md
+++ b/packages/docs/docs/api/cli.md
@@ -54,6 +54,10 @@ Start development server in debug mode.
 
 Start development server in silent mode.
 
+#### --max-concurrency
+
+Set the max concurrency for rendering pages, to prevent OOM when rendering massive docs.
+
 ### dev
 
 Start a development server. All options from `vuepress build` are available. And there are several options specifically for dev:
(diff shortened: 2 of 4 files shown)


troyeagle:
### Why not p-limit
I did not dig through it, but in my case, it seems p-limit will cause the whole process turn into serial, which is actually much slower.
manual
pLimit

packages/@vuepress/core/lib/node/build/index.js
```diff
@@ -90,10 +91,16 @@ module.exports = class Build extends EventEmitter {
...
-
-    const pagePaths = await Promise.all(
-      this.context.pages.map(page => this.renderPage(page))
-    )
+    if (this.maxConcurrency) logger.info(`max concurrency set: ${this.maxConcurrency}`)
+
+    const pagePaths = []
+    while (this.context.pages.length) {
+      const segmentPaths = await Promise.all(
+        this.context.pages.splice(0, this.maxConcurrency || 100000)
```
ulivz, line 99: It is not advisable to modify `context.pages` directly

ulivz (changes requested):
Could you address the comment and update the documentation?

packages/@vuepress/core/lib/node/build/index.js
troyeagle: changed to slice + for loop

ulivz (approved):

sr2ds:
Hello, @troyeagle first congrats by this job, its really cool.

I'm trying to optimize this flow or trying understand better.
I have a blog with >2K posts and the memory use on build is so big.
I'm trying setting up some different max concurrency but without success yet.

I don't know if do you remember but, on my tests, the memory is never cleanuped while the loop run.
I'm trying cleanup the memory while the process is running but I can´t do it.
In other words, I can to limit the async process with the max concurrency, but the memory usage is more big each time.
This is a dump of memory after each promise call.

image

My tests there are around this place:
image
I tried set as null the variables to 'clean-up' memory, but I guess the GC do it alone. BUT if do, is not doing.

Thanks for your attention, if you have some idea to share with me, will be great.

troyeagle:
> 

As I remember, what magic happens may be in the method `this.renderPage`, or your `processAsyncRenderPage` which uses `vue-server-renderer` and use its `renderString` utility. 

The server-renderer may be stateful, it not just accepts an object and uses template to parse it, but [run with a context](https://github.com/vuejs/vue/blob/8d3fce029f20a73d5d0b1ff10cbf6fa73c989e62/packages/vue-server-renderer/build.dev.js#L9577) and then create a bunch of async rendering tasks. The context may [expand during the rendering process](https://github.com/vuejs/vue/blob/8d3fce029f20a73d5d0b1ff10cbf6fa73c989e62/packages/vue-server-renderer/build.dev.js#L9117) to save building cache for optimizing following tasks' rendering performance.

I think it's not the results' not being GC-ed but the context growing that causes your memory growing. It seems vuepress always uses one context, when a group of tasks ends, just some of the spaces are set free. @sr2ds You could check **whether** the total size of 5 pages' html output string is the same as your memory growth per time, or use `chrome://inspect` to see what object in the memory is really growing as the task goes on, then you may have idea of what objects or attributes you should set null to trigger garbage collection. 

In all, the memory problem is the tradeoff of server-side rendering optimization. Vuepress maybe did not expect such huge size of pages, and in `vuepress build` case, building speed may be the first concern.

sr2ds:
Thanks so much for your time and answer.

To be honest I don't know how do a good review on inspect mode, but looks that the `string` type of data is the most used, I don't sure if can be the returns of `renderString` stored on `context`.

I made some tests but without success yet.

I saw this other [issue](https://github.com/vuepress/vuepress-next/issues/994) with some comparatives between vuepress and vitepress, I'll try a little bit more but looks more prudent to migrate to `vitepres` on this specific blog case.

